PR #3, opened 2015-03-03
Chat: Adapt chatrooms event page with new layout

nop33:


pferreir:
Already merged in 1c37e49.

PR #5, opened 2015-03-27
Fix move vidyo deps out of vc module

jacquesd:
- moves the remaining Vidyo dependent code out of the VC module and into the `vc_vidyo` plugin
- updates the `vc_dummy` plugin to work with the changes in the VC module for Indico `1.9.2`

> **WARNING**: This should only be merged jointly with the PR https://github.com/indico/indico/pull/1816 to avoid breaking the VC plugins.


pferreir:
"Link" widget kind of broken...

dummy plugin

ThiefMaster:
Merged in b462f14

PR #6, opened 2015-04-09
Update owner when the user is merged

nop33:
An error report led to find out that when a user is merged and before he was an owner of a VC room, the `vc_room.data['owner']` was not updated properly.


nop33:
It shouldn't be merged yet.

@ThiefMaster suggested to use `ext.vc_room` instead of querying for a vc_room (which makes sense).
However the following code does not flush the new data in the database:
```python
for ext in VidyoExtension.find(owned_by_id=old_id):
    ext.owned_by_id = new_id
    ext.vc_room.data['owner'] = principal_to_tuple(user)
    flag_modified(ext.vc_room, 'data')
    db.session.flush()
```

any suggestion?

ThiefMaster:
Merged in 44acb37

PR #23, opened 2015-12-09
Jupyter: added default sane MathJax config

pferreir:
Won't work without unsafe-eval CSP :/


pferreir:
@ThiefMaster good point. Done.

PR #27, opened 2016-04-12
Adapt importer to new timetable

florv:


ThiefMaster:
Can you please rebase and re-open this against master? (probably needs some more testing too, but I think that makes more sense after the legacy code cleanup is done)

PR #32, opened 2016-07-21
LiveSync - link to categories and sessions

pferreir:
Simplified the way records are processed.

See also https://github.com/indico/indico/pull/2545.

Missing:
* [x] Unit tests (need some dummy DB objects)
* [x] Checking that Invenio/CERN plugins work fine
* [x] Some extra smoke testing (handling deletes, etc...)

### DIFF
--- a/livesync/indico_livesync/cli.py
+++ b/livesync/indico_livesync/cli.py
@@ -26,6 +26,7 @@ from indico.core.db.sqlalchemy.util.session import update_session_options
 from indico.modules.events.models.events import Event
 from indico.util.console import cformat
 
+
 from indico_livesync.models.agents import LiveSyncAgent
 
 
@@ -66,7 +67,7 @@ def agents():
 
 
 @cli_manager.option('agent_id')
-@cli_manager.option('--force', action='store_true')
+@cli_manager.option('--force', action='store_true', help="Perform export even if it has already been done once.")
 def initial_export(agent_id, force=False):
     """Performs the initial data export for an agent"""
     update_session_options(db)
--- a/livesync/indico_livesync/cli.py
+++ b/livesync/indico_livesync/cli.py
@@ -23,8 +23,8 @@ from terminaltables import AsciiTable
 
 from indico.core.db import db, DBMgr
 from indico.core.db.sqlalchemy.util.session import update_session_options
-from indico.util.console import cformat, conferenceHolderIterator
-from MaKaC.conference import ConferenceHolder
+from indico.modules.events.models.events import Event
+from indico.util.console import cformat
 
 from indico_livesync.models.agents import LiveSyncAgent
 
@@ -83,22 +83,13 @@ def initial_export(agent_id, force=False):
         print cformat('To re-run it, use %{yellow!}--force%{reset}')
         return
 
-    def _iter_events():
-        for i, (_, event) in enumerate(conferenceHolderIterator(ConferenceHolder(), deepness='event'), 1):
-            yield event
-            if i % 1000 == 0:
-                # Clean local ZEO cache
-                transaction.abort()
-
-    with DBMgr.getInstance().global_connection():
-        agent.create_backend().run_initial_export(_iter_events())
-
+    agent.create_backend().run_initial_export(Event.find(is_deleted=False))
     agent.initial_data_exported = True
     db.session.commit()
 
 
 @cli_manager.option('agent_id')
-@cli_manager.option('--force', action='store_true')
+@cli_manager.option('--force', action='store_true', help="Run even if initial export was not done")
 def run(agent_id, force=False):
     """Runs the livesync agent"""
     update_session_options(db)
--- a/livesync/indico_livesync/handler.py
+++ b/livesync/indico_livesync/handler.py
@@ -165,7 +165,7 @@ def _note_changed(note, **kwargs):
 def _attachment_changed(attachment_or_folder, **kwargs):
     folder = getattr(attachment_or_folder, 'folder', attachment_or_folder)
     if folder.link_type not in (LinkType.category, LinkType.session):
-        _register_change(folder.object, ChangeType.data_changed)
+        _register_change(folder.object.event_new, ChangeType.data_changed)
 
 
 def _apply_changes(sender, **kwargs):
--- a/livesync/indico_livesync/handler.py
+++ b/livesync/indico_livesync/handler.py
@@ -24,15 +24,16 @@ from sqlalchemy import inspect
 from indico.core import signals
 from indico.core.db.sqlalchemy.links import LinkType
 from indico.core.db.sqlalchemy.protection import ProtectionMode
+from indico.modules.categories.models.categories import Category
 from indico.modules.events import Event
 from indico.modules.events.contributions.models.contributions import Contribution
 from indico.modules.events.contributions.models.subcontributions import SubContribution
 from indico.modules.events.sessions import Session
 from indico.util.event import unify_event_args
-from MaKaC.conference import Conference
+
 
 from indico_livesync.models.queue import LiveSyncQueueEntry, ChangeType
-from indico_livesync.util import obj_ref, is_ref_excluded
+from indico_livesync.util import obj_ref, get_excluded_categories
 
 
 def connect_signals(plugin):
@@ -40,8 +41,8 @@ def connect_signals(plugin):
     plugin.connect(signals.after_process, _apply_changes)
     plugin.connect(signals.before_retry, _clear_changes)
     # moved
-    # plugin.connect(signals.category.moved, _moved)
-    # plugin.connect(signals.event.moved, _moved)
+    plugin.connect(signals.category.moved, _moved)
+    plugin.connect(signals.event.moved, _moved)
     # created
     plugin.connect(signals.event.created, _created)
     plugin.connect(signals.event.contribution_created, _created)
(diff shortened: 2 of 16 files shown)


livesync/indico_livesync/handler.py
```diff
@@ -88,12 +79,17 @@ def connect_signals(plugin):
     plugin.connect(signals.attachments.attachment_updated, _attachment_changed)
 
 
-def _moved(obj, old_parent, new_parent, **kwargs):
+def _moved_event(obj, old_category, **kwargs):
```
OmeGak: It's not clear what this function does unless you go into `_moved`. What about a more descriptive name?

ThiefMaster: Does it really matter though? It's a signal handler, not some API.
Anyway, It looks like the signal could be connected directly to `_moved`

livesync/indico_livesync/simplify.py
```diff
@@ -16,10 +16,17 @@
 
 from __future__ import unicode_literals
 
+import itertools
 from collections import defaultdict
+from sqlalchemy.orm import joinedload
```
ThiefMaster: that's not stdlib

livesync/indico_livesync/models/queue.py
```diff
@@ -176,62 +209,72 @@ class LiveSyncQueueEntry(db.Model):
...
-        for agent in LiveSyncAgent.find():
-            for change in changes:
+        for change in changes:
+            obj = obj_deref(ref)
+            if isinstance(obj, Category):
+                if excluded_categories & {c.id for c in obj.chain_query}:
+                    continue
+            else:
+                event = obj if isinstance(obj, Event) else obj.event_new
+                if Event.query.filter(
```
OmeGak: Mixed multi-line style

livesync/indico_livesync/handler.py
OmeGak: I know it's a handler, but I think it's generally better to name the functions after what they do and not after what calls them.

ThiefMaster: That does not make much sense in livesync - all of them eventually create livesync entries for an action

OmeGak: `_register_event_move` and nobody has to wonder.

pferreir: I'd say that since the convention with event handlers right now is to name the function after the event, we should keep this one consistent.
In this particular case, it would have to be `_register_move`, since this handler is also used by categories. Anyway, prefixing everything with `_register` seems like a bit of an overkill to me...

livesync/indico_livesync/simplify.py
```diff
@@ -44,23 +53,80 @@ def process_records(records):
...
+    category_prot_records -= category_move_records  # A move already implies sending the whole record
+
+    # Protection changes are handled differently, as there may not be the need to re-generate the record
+    if category_prot_records:
+        for categ in Category.find(Category.id.in_(category_prot_records)):
+            cte = categ.get_protection_parent_cte()
+            # Update only children that inherit
+            inheriting_categ_children = (Event.query
+                                         .join(Category)
+                                         .join(cte, (Category.id == cte.c.id) & (cte.c.protection_parent == categ.id)))
```
ThiefMaster: Do you actually need to `.join(Category)` here? Why not use `Event.category_id` in this filter instead

livesync/indico_livesync/util.py
```diff
@@ -46,36 +47,24 @@ def obj_ref(obj):
...
+@memoize_request
 def obj_deref(ref):
     """Returns the object identified by `ref`"""
     from indico_livesync.models.queue import EntryType
     if ref['type'] == EntryType.category:
-        return CategoryManager().getById(ref['category_id'], True)
+        return Category.get(ref['category_id'])
     elif ref['type'] == EntryType.event:
-        return Event.get(ref['event_id'])
+        return Event.find(id=ref['event_id']).one()
```
ThiefMaster: `Event.get_one()`?

livesync/indico_livesync/util.py
```diff
@@ -46,36 +47,24 @@ def obj_ref(obj):
     return ImmutableDict(ref)
 
 
+@memoize_request
 def obj_deref(ref):
     """Returns the object identified by `ref`"""
     from indico_livesync.models.queue import EntryType
     if ref['type'] == EntryType.category:
-        return CategoryManager().getById(ref['category_id'], True)
+        return Category.get(ref['category_id'])
```
ThiefMaster: `get_one` for consistency

livesync/indico_livesync/util.py
```diff
@@ -46,36 +47,24 @@ def obj_ref(obj):
...
     elif ref['type'] == EntryType.event:
-        return Event.get(ref['event_id'])
+        return Event.find(id=ref['event_id']).one()
+    elif ref['type'] == EntryType.session:
+        return Session.find(id=ref['session_id']).one()
     elif ref['type'] == EntryType.contribution:
-        return Contribution.get(ref['contrib_id'])
+        return Contribution.find(id=ref['contrib_id']).one()
```
ThiefMaster: `get_one()`

livesync/indico_livesync/util.py
```diff
@@ -91,25 +80,18 @@ def clean_old_entries():
...
-    from indico_livesync.models.queue import EntryType
-    if ref['type'] == EntryType.category:
-        return ref['category_id'] in get_excluded_categories()
-    else:
-        obj = obj_deref(ref)
-        return unicode(obj.event_new.category_id) in {unicode(x) for x in get_excluded_categories()}
+def compound_id(obj):
+    """Generate a hierarchical compound ID, separated by dots."""
+    if isinstance(obj, (Category, Session)):
+        raise ValueError('Compound IDs are not supported for this entry type')
```
ThiefMaster: `TypeError`?

livesync/tests/queue_test.py
```diff
@@ -0,0 +1,57 @@
...
+}
+
+
+def test_excluded_categories(mocker, monkeypatch, db, create_category, dummy_agent):
+    """Test if category exclusions work"""
+
+    # IMPORTANT: `dummy_agent` is actually needed here, since ``LiveSyncQueueEntry.create``
+    # will look for all existing agents in the DB. `db.session.add` is not necessary but
+    # it keeps static checkers happy.
+    db.session.add(dummy_agent)
```
ThiefMaster: You can use `@pytest.mark.usefixtures('dummy_agent')` if you just need to enable a fixture but not actually access it

livesync/indico_livesync/cli.py
```diff
@@ -26,6 +26,7 @@
 from indico.modules.events.modes.events import Event
 from indico.util.console import cformat
 
+
```
ThiefMaster: too much whitespace

livesync/indico_livesync/uploader.py
```diff
@@ -87,7 +87,6 @@ def processed_records(self, records):
             self.logger.debug('Marking as processed: %s', record)
             record.processed = True
         db.session.commit()
-        transaction.abort()  # clear ZEO cache
```
ThiefMaster: Is the commit message correct? This doesn't seem to be related to livesync-invenio

livesync/indico_livesync/handler.py
```diff
@@ -23,31 +23,31 @@
...
 
 def connect_signals(plugin):
     # request
     plugin.connect(signals.after_process, _apply_changes)
     plugin.connect(signals.before_retry, _clear_changes)
     # moved
-    # plugin.connect(signals.category.moved, _moved)
-    # plugin.connect(signals.event.moved, _moved)
+    plugin.connect(signals.category.moved, _moved)
+    plugin.connect(signals.event.moved, _moved_event)
```
ThiefMaster: `_event_moved` to be consistent with `_event_deleted`

livesync/indico_livesync/handler.py
```diff
@@ -110,6 +107,10 @@ def _created(obj, **kwargs):
     _register_change(obj, ChangeType.created)
 
 
+def _event_deleted(obj, **kwargs):
```
ThiefMaster: I'd rename the first argument to `event` and decorate it with `@unify_event_args`. Less code to change once we change the signal to pass the new Event object (which we could do already - the only case where the legacy object is needed is the signal handler that logs the ZODB OID, and that could simply be moved into the Conference.delete method).

Actually, you could even use `_deleted` for the event deleted signal and add the decorator there.

livesync/indico_livesync/handler.py
```diff
@@ -165,16 +137,14 @@ def _note_changed(note, **kwargs):
...
 
 
 def _apply_changes(sender, **kwargs):
     if not hasattr(g, 'livesync_changes'):
         return
     for ref, changes in g.livesync_changes.iteritems():
-        if is_ref_excluded(ref):
-            continue
-        LiveSyncQueueEntry.create(changes, ref)
+        LiveSyncQueueEntry.create(changes, ref, excluded_categories=get_excluded_categories())
```
ThiefMaster: does it really make sense to call it inside the loop? i know it's memoized, but it still seems cleaner to do it outside

livesync/indico_livesync/migrations/201607151736_205f944640f6_add_session_to_livesyncqueueentry.py
```diff
@@ -0,0 +1,84 @@
...
+down_revision = '230c086da074'
+
+
+def upgrade():
+    op.add_column('queues', sa.Column('session_id', sa.Integer(), nullable=True), schema='plugin_livesync')
+    op.create_foreign_key(None,
+                          'queues', 'sessions',
+                          ['session_id'], ['id'],
+                          source_schema='plugin_livesync', referent_schema='events')
+    op.create_index(op.f('ix_queues_session_id'), 'queues', ['session_id'], unique=False, schema='plugin_livesync')
```
ThiefMaster: first argument can be `None`; the name will be generated automatically

livesync/indico_livesync/util.py
pferreir: I think you originally wrote that some years ago 😉  But yeah, makes more sense.

livesync/tests/queue_test.py
pferreir: Yeah, I didn't think of that. I had seen it before.

livesync/indico_livesync/handler.py
pferreir: Jolly good idea! 👍

livesync/indico_livesync/handler.py
pferreir: You mean `get_excluded_categories()`? No it doesn't.

ThiefMaster: yes, that one

livesync/indico_livesync/models/queue.py
```diff
@@ -175,62 +209,61 @@ class LiveSyncQueueEntry(db.Model):
...
-            for change in changes:
+        obj = obj_deref(ref)
+
+        if isinstance(obj, Category):
+            if excluded_categories & {c.id for c in obj.chain_query}:
+                return
+        else:
+            event = obj if isinstance(obj, Event) else obj.event_new
+            if (Event.query
+                    .filter(Event.id == event.id, Event.category_chain_overlaps(excluded_categories))
```
ThiefMaster: wrong indentation

livesync/indico_livesync/models/queue.py
```diff
@@ -175,62 +209,61 @@ class LiveSyncQueueEntry(db.Model):
...
+        :param excluded_categories: set of categories (IDs) whose items
+                                    will not be tracked
         """
         ref = dict(ref)
-        for agent in LiveSyncAgent.find():
-            for change in changes:
+        obj = obj_deref(ref)
+
+        if isinstance(obj, Category):
+            if excluded_categories & {c.id for c in obj.chain_query}:
```
ThiefMaster: `any(c.id in excluded_categories for c in obj.chain_query)`

PR #36, opened 2016-09-12
Move forms from page to dialog

DavidAndreev:
Part of issue: indico/indico#2596

### DIFF
--- a/chat/indico_chat/controllers/management.py
+++ b/chat/indico_chat/controllers/management.py
@@ -17,7 +17,7 @@
 from __future__ import unicode_literals
 
 from flask import session, flash, redirect, jsonify
-from flask_pluginengine import current_plugin
+from flask_pluginengine import current_plugin, render_plugin_template
 
 from indico.core.db import db
 from indico.core.db.sqlalchemy.util.models import attrs_changed
@@ -27,6 +27,7 @@ from indico.modules.events.logs import EventLogRealm, EventLogKind
 from indico.util.date_time import now_utc
 from indico.util.string import to_unicode
 from indico.web.forms.base import FormDefaults
+from indico.web.util import jsonify_data, jsonify_template
 
 from indico_chat import _
 from indico_chat.controllers.base import RHChatManageEventBase, RHEventChatroomMixin
@@ -82,10 +83,9 @@ class RHChatManageEventModify(RHEventChatroomMixin, RHChatManageEventBase):
             flash(_('Chatroom updated'), 'success')
             self.event.log(EventLogRealm.management, EventLogKind.change, 'Chat',
                            'Chatroom updated: {}'.format(self.chatroom.name), session.user)
-            return redirect(url_for_plugin('.manage_rooms', self.event))
-        return WPChatEventMgmt.render_template('manage_event_edit.html', self._conf, form=form,
-                                               event_chatroom=self.event_chatroom, event=self.event)
-
+            return jsonify_data(flash=False)
+        return jsonify_template('manage_event_edit.html', render_plugin_template, form=form,
+                                event=self.event, event_chatroom=self.event_chatroom)
 
 class RHChatManageEventRefresh(RHEventChatroomMixin, RHChatManageEventBase):
     """Synchronizes the local chatroom data with the XMPP server"""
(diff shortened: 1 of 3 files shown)


chat/indico_chat/controllers/management.py
```diff
@@ -17,6 +17,7 @@
 from __future__ import unicode_literals
 
 from flask import session, flash, redirect, jsonify
+from flask.ext.pluginengine import render_plugin_template
```
ThiefMaster: This is deprecated, use `flask_pluginengine` instead. Pisses me off that the PyCharm devs still didn't fix this.. https://youtrack.jetbrains.com/issue/PY-13974

chat/indico_chat/templates/manage_event.html
```diff
@@ -27,7 +27,14 @@
                                         {% if logs_enabled and not chatroom.custom_server %}
                                             <a href="{{ url_for_plugin('.manage_rooms_logs', event_chatroom) }}" class="i-button icon-file-content" title="{% trans %}Logs{% endtrans %}"></a>
                                         {% endif %}
-                                        <a href="{{ url_for_plugin('.manage_rooms_modify', event_chatroom) }}" class="i-button icon-edit" title="{%- trans %}Edit{% endtrans -%}"></a>
+                                        <a href="#" class="i-button icon-edit"
+                                           data-href="{{ url_for_plugin('.manage_rooms_modify', event_chatroom) }}"
+                                           data-title="{%- trans %}Edit chatroom{% endtrans -%}"
+                                           data-ajax-dialog
+                                           data-hide-page-header
```
ThiefMaster: this is not needed if you adapt the template properly (remove inheritance and the blocks)

chat/indico_chat/templates/manage_event.html
```diff
@@ -27,7 +25,14 @@
                                         {% if logs_enabled and not chatroom.custom_server %}
                                             <a href="{{ url_for_plugin('.manage_rooms_logs', event_chatroom) }}" class="i-button icon-file-content" title="{% trans %}Logs{% endtrans %}"></a>
                                         {% endif %}
-                                        <a href="{{ url_for_plugin('.manage_rooms_modify', event_chatroom) }}" class="i-button icon-edit" title="{%- trans %}Edit{% endtrans -%}"></a>
+                                        <a href="#" class="i-button icon-edit"
+                                           data-href="{{ url_for_plugin('.manage_rooms_modify', event_chatroom) }}"
+                                           data-title="{%- trans %}Edit chatroom{% endtrans -%}"
```
nop33: The `-` inside `{%- trans %}` (and `{% endtrans -%}`) is not needed. It's only used to remove empty space and in this case there's none :wink:

chat/indico_chat/templates/manage_event.html
```diff
@@ -1,7 +1,5 @@
-{% extends 'layout/base.html' %}
 {% from 'forms/_form.html' import form_header, form_row, form_footer  %}
 
-{% block title %}{% trans %}Chat Rooms{% endtrans %}{% endblock %}
 {% block content %}
```
ThiefMaster: While not absolutely necessary (the block simply doesn't do anything if no inheritance is used), it would be cleaner to remove this block tag too and outdent its contents

chat/indico_chat/templates/manage_event.html
nop33: Same for the other cases too. It would be useful tho in the line
```
{% trans %}Create new room{% endtrans %}
``` 
:smiley:

chat/indico_chat/templates/manage_event.html
```diff
@@ -27,7 +26,13 @@
                                         {% if logs_enabled and not chatroom.custom_server %}
                                             <a href="{{ url_for_plugin('.manage_rooms_logs', event_chatroom) }}" class="i-button icon-file-content" title="{% trans %}Logs{% endtrans %}"></a>
                                         {% endif %}
-                                        <a href="{{ url_for_plugin('.manage_rooms_modify', event_chatroom) }}" class="i-button icon-edit" title="{%- trans %}Edit{% endtrans -%}"></a>
+                                        <a href="#" class="i-button icon-edit"
+                                           data-href="{{ url_for_plugin('.manage_rooms_modify', event_chatroom) }}"
+                                           data-title="{% trans %}Edit chatroom{% endtrans %}"
+                                           data-ajax-dialog
+                                           data-reload-after
+                                           title="{%- trans %}Edit{% endtrans -%}">
```
OmeGak: Here as well, we don't need to trim white space. This will suffice:
`{% trans %}Edit{% endtrans %}`

You can read about it here:
http://jinja.pocoo.org/docs/dev/templates/#whitespace-control

ThiefMaster: It doesn't harm though ;) But yeah, I wouldn't add it in cases where it makes no difference either

ThiefMaster: Like before, I'd put this further up to have the data attributes at the end (and the ones with no value at the very end)

PR #37, opened 2016-09-14
retrieving right room from WSDL Vidyo web service

gasparuben:
Further filter results from a call to Vidyo Admin web service in order to find a room.


vc_vidyo/indico_vc_vidyo/api/client.py
```diff
@@ -79,9 +79,28 @@ def create_room_object(self, **kwargs):
     def find_room(self, extension):
         filter_ = self.client.factory.create('Filter')
         filter_.query = extension
+        filter_.limit = 40
         filter_.dir = 'DESC'
 
-        return self.soap.getRooms(filter_).room
+        target_room = None
+        counter = 0
+        from indico_vc_vidyo.plugin import VidyoPlugin
```
ThiefMaster: local imports should be at the top of the function

vc_vidyo/indico_vc_vidyo/api/client.py
```diff
@@ -79,9 +79,28 @@ def create_room_object(self, **kwargs):
...
+        while True:
+            filter_.start = (counter*filter_.limit)
```
ThiefMaster: no need for the parentheses, and there should be one space on each side of the `*`

vc_vidyo/indico_vc_vidyo/api/client.py
```diff
@@ -79,9 +79,28 @@ def create_room_object(self, **kwargs):
...
+            VidyoPlugin.logger.warning('filter is: %s', filter_.start)
+            rooms = self.soap.getRooms(filter_).room
+            for room in rooms:
+                if (int(room.extension) == int(extension)):
```
OmeGak: No need for parenthesis in `if` statements.

vc_vidyo/indico_vc_vidyo/api/client.py
```diff
@@ -79,9 +79,28 @@ def create_room_object(self, **kwargs):
...
+        filter_.limit = 40
         filter_.dir = 'DESC'
 
-        return self.soap.getRooms(filter_).room
+        target_room = None
+        counter = 0
+        from indico_vc_vidyo.plugin import VidyoPlugin
+        while True:
+            filter_.start = (counter*filter_.limit)
+            VidyoPlugin.logger.warning('filter is: %s', filter_.start)
```
ThiefMaster: I don't think it makes sense to keep this logging here - *especially* not with a level higher than `debug` (but i think it's not that important at all to log this)

vc_vidyo/indico_vc_vidyo/api/client.py
ThiefMaster: don't put parentheses around a single-line criterion

vc_vidyo/indico_vc_vidyo/api/client.py
```diff
@@ -79,9 +79,28 @@ def create_room_object(self, **kwargs):
...
+            rooms = self.soap.getRooms(filter_).room
+            for room in rooms:
+                if (int(room.extension) == int(extension)):
+                    VidyoPlugin.logger.info('Room: %s has been found.', room)
+                    target_room = room
+                    break
+                else:
+                    VidyoPlugin.logger.info('Dismissing room extension %s', room.extension)
+            if not target_room:
```
ThiefMaster: I think this would be clearer:
```
if target_room:
    return target_room
counter += 1
```

vc_vidyo/indico_vc_vidyo/api/client.py
OmeGak: Missing whitespace around the `*` operator as per [PEP8](https://www.python.org/dev/peps/pep-0008/).

vc_vidyo/indico_vc_vidyo/plugin.py
```diff
@@ -203,8 +203,10 @@ def create_room(self, vc_room, event):
 
             else:
                 # get room back, in order to fetch Vidyo-set parameters
-                created_room = client.find_room(extension)[0]
+                created_room = client.find_room(extension)
 
+                if not created_room:
+                    raise VCRoomNotFoundError(_("This room doesnt exist in Vidyo"))
```
ThiefMaster: `doesn't` or `does not`

Also, I think the error message is not really clear. I'd use something like "Could not find newly created room in Vidyo".

vc_vidyo/indico_vc_vidyo/api/client.py
OmeGak: We group imports at the top of the file unless they cause a circular import. In that case, we put them as the first lines of the function.

ThiefMaster: I'm quite sure it'd cause a circular import here - but I don't think it's needed here at all, the log messages are not really useful except when testing the code locally

vc_vidyo/indico_vc_vidyo/api/client.py
```diff
@@ -77,11 +77,28 @@ def create_room_object(self, **kwargs):
...
+        target_room = None
+        counter = 0
+
+        while True:
+            filter_.start = counter*filter_.limit
+            rooms = self.soap.getRooms(filter_).room
+            for room in rooms:
+                if int(room.extension) == int(extension):
+                    VidyoPlugin.logger.debug('Room: %s has been found.', room)
+                    target_room = room
```
OmeGak: Why not returning `room` from here directly? That way you can get rid of `break` and `if target_room` down in `L99`.

vc_vidyo/indico_vc_vidyo/api/client.py
```diff
@@ -77,11 +77,28 @@ def create_room_object(self, **kwargs):
...
         filter_ = self.client.factory.create('Filter')
         filter_.query = extension
+        filter_.limit = 40
         filter_.dir = 'DESC'
 
-        return self.soap.getRooms(filter_).room
+        target_room = None
+        counter = 0
+
+        while True:
```
OmeGak: What happens if `self.soap.getRooms(filter_).room` doesn't return any room for some reason? Then we'd be stuck in an infinite loop. Is this really what we want?

gasparuben: Yes, I got the API doc from Vidyo, I have improved that.

vc_vidyo/indico_vc_vidyo/api/client.py
```diff
@@ -77,11 +77,26 @@ def create_room_object(self, **kwargs):
...
+        filter_.limit = 40
         filter_.dir = 'DESC'
 
-        return self.soap.getRooms(filter_).room
+        counter = 0
+
+        while True:
+            filter_.start = counter*filter_.limit
+            response = self.soap.getRooms(filter_)
+            if response.total == 0:
```
ThiefMaster: I would go for `if not response.total:`

vc_vidyo/indico_vc_vidyo/api/client.py
```diff
@@ -77,11 +77,26 @@ def create_room_object(self, **kwargs):
...
         filter_ = self.client.factory.create('Filter')
         filter_.query = extension
+        filter_.limit = 40
         filter_.dir = 'DESC'
 
-        return self.soap.getRooms(filter_).room
+        counter = 0
+
+        while True:
+            filter_.start = counter*filter_.limit
```
ThiefMaster: I know PEP8 leaves this unspecified but for the sake of consistency I'd still put spaces around the `*` - especially since in this case not using spaces does not improve readability (since it's only two operands)

pferreir:
> LGTM besides the two things I commented

Then we can merge as soon as @gasparuben pushes an updated version (if he hasn't yet).

gasparuben:
Yes, it's done and tested.

PR #39, opened 2016-12-07
Vc vidyo removing obsolete vc rooms

gasparuben:
After testing with Rene and had agreed format of the output.

### DIFF
--- a/vc_vidyo/indico_vc_vidyo/http_api.py
+++ b/vc_vidyo/indico_vc_vidyo/http_api.py
@@ -1,0 +1,65 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2016 European Organization for Nuclear Research (CERN).
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 3 of the
+# License, or (at your option) any later version.
+#
+# Indico is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from flask import request
+
+from indico.modules.vc.models.vc_rooms import VCRoom, VCRoomStatus
+from indico.web.http_api.hooks.base import HTTPAPIHook
+
+
+class DeleteVCRoomAPI(HTTPAPIHook):
+    PREFIX = 'api'
+    TYPES = ('deletevcroom',)
+    RE = r'vidyo'
+    GUEST_ALLOWED = False
+    VALID_FORMATS = ('json',)
+    COMMIT = True
+    HTTP_POST = True
+
+    def _hasAccess(self, aw):
+        from indico_vc_vidyo.plugin import VidyoPlugin
+        return aw.getUser().user in VidyoPlugin.settings.acls.get('managers')
+
+    def _getParams(self):
+        super(DeleteVCRoomAPI, self)._getParams()
+        self._room_ids = map(int, request.form.getlist('rid'))
+
+    def api_deletevcroom(self, aw):
+        from indico_vc_vidyo.plugin import VidyoPlugin
+        from indico_vc_vidyo.api import APIException
+
+        success = []
+        failed = []
+        not_in_db = []
+
+        for rid in self._room_ids:
+            room = VCRoom.query.filter(VCRoom.type == 'vidyo',
+                                       VCRoom.status == VCRoomStatus.created,
+                                       VCRoom.data.contains({'vidyo_id': str(rid)})).first()
+            if not room:
+                not_in_db.append(rid)
+                continue
+            try:
+                room.plugin.delete_room(room, None)
+            except APIException:
+                failed.append(rid)
+                VidyoPlugin.logger.exception('Could not delete VC room %s', room)
+            else:
+                room.status = VCRoomStatus.deleted
+                success.append(rid)
+                VidyoPlugin.logger.info('%s deleted', room)
+
+        return {'success': success, 'failed': failed, 'missing': not_in_db}
--- a/vc_vidyo/indico_vc_vidyo/plugin.py
+++ b/vc_vidyo/indico_vc_vidyo/plugin.py
@@ -33,12 +33,14 @@ from indico.modules.vc.views import WPVCEventPage, WPVCManageEvent
 from indico.util.user import retrieve_principal
 from indico.web.forms.fields import IndicoPasswordField
 from indico.web.forms.widgets import CKEditorWidget
+from indico.web.http_api.hooks.base import HTTPAPIHook
 
 from indico_vc_vidyo import _
 from indico_vc_vidyo.api import AdminClient, APIException, RoomNotFoundAPIException
 from indico_vc_vidyo.blueprint import blueprint
 from indico_vc_vidyo.cli import cli_manager
 from indico_vc_vidyo.forms import VCRoomForm, VCRoomAttachForm
+from indico_vc_vidyo.http_api import DeleteVCRoomAPI
 from indico_vc_vidyo.util import iter_user_identities, iter_extensions, update_room_from_obj
 from indico_vc_vidyo.models.vidyo_extensions import VidyoExtension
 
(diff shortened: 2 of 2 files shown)


vc_vidyo/indico_vc_vidyo/http_api.py
```diff
@@ -0,0 +1,73 @@
...
+                break
+            room = VCRoom.query.filter(VCRoom.type == 'vidyo',
+                                       VCRoom.status == VCRoomStatus.created,
+                                       VCRoom.data.contains({'vidyo_id': str(id)})).first()
+            if not room:
+                notexist_at_indico.append(id)
+                continue
+            try:
+                room.plugin.delete_room(room, None)
+            except:
```
ThiefMaster: catch the actual exception you expect, or at least `Exception` - bare `except:` is never a good idea since it catches things like `KeyboardInterrupt` or `SystemExit` too.

vc_vidyo/indico_vc_vidyo/http_api.py
```diff
@@ -0,0 +1,73 @@
...
+        self._room_ids = request.form.getlist('room_id')
+
+    def api_deletevcroom(self, aw):
+        from indico_vc_vidyo.plugin import VidyoPlugin
+        success = []
+        failed = []
+        notexist_at_indico = []
+
+        counter = 0
+        for id in self._room_ids:
```
ThiefMaster: `room_id` as `id` is the name of a builtin function

vc_vidyo/indico_vc_vidyo/http_api.py
```diff
@@ -0,0 +1,73 @@
...
+    # Just let the first 10 to be deleted
+    MAX_COUNT = 10
+
+    def _hasAccess(self, aw):
+        from indico_vc_vidyo.plugin import VidyoPlugin
+        return aw.getUser().user in VidyoPlugin.settings.acls.get('managers')
+
+    def _getParams(self):
+        super(VCRoomCleanUpAPI, self)._getParams()
+        self._room_ids = request.form.getlist('room_id')
```
ThiefMaster: I would use `self._room_ids = map(int, request.form.getlist('room_id'))` to ensure they are all ints

vc_vidyo/indico_vc_vidyo/http_api.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def api_deletevcroom(self, aw):
+        from indico_vc_vidyo.plugin import VidyoPlugin
+        success = []
+        failed = []
+        notexist_at_indico = []
```
ThiefMaster: `missing` or `not_in_db`?

vc_vidyo/indico_vc_vidyo/http_api.py
```diff
@@ -0,0 +1,73 @@
...
+class VCRoomCleanUpAPI(HTTPAPIHook):
+    PREFIX = 'api'
+    TYPES = ('deletevcroom', )
+    RE = r'vidyo'
+    GUEST_ALLOWED = False
+    VALID_FORMATS = ('json', )
+    COMMIT = True
+    HTTP_POST = True
+
+    # Just let the first 10 to be deleted
```
ThiefMaster: is that useful? seems to be an arbitrary restriction - a malicious client could simply send multiple requests anyway...

vc_vidyo/indico_vc_vidyo/http_api.py
```diff
@@ -0,0 +1,73 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from flask import request
+
+from indico.web.http_api.hooks.base import HTTPAPIHook
+from indico.modules.vc.models.vc_rooms import VCRoom, VCRoomStatus
+
+
+class VCRoomCleanUpAPI(HTTPAPIHook):
```
ThiefMaster: `DeleteVCRoomAPI`?

vc_vidyo/indico_vc_vidyo/http_api.py
```diff
@@ -0,0 +1,73 @@
...
+    PREFIX = 'api'
+    TYPES = ('deletevcroom', )
```
ThiefMaster: usually people don't put the space between `,)` (since there's no other element following anyawy)

vc_vidyo/indico_vc_vidyo/http_api.py
```diff
@@ -0,0 +1,73 @@
...
+            else:
+                counter += 1
+                room.status = VCRoomStatus.deleted
+                success.append(room.data['vidyo_id'])
+                VidyoPlugin.logger.info('{} deleted', room)
+
+        return {'success': '{}'.format(success),
+                'failed': '{}'.format(failed),
+                'notexist_at_indico': '{}'.format(notexist_at_indico)
+                }
```
ThiefMaster: should be at the end of the previous line

vc_vidyo/indico_vc_vidyo/http_api.py
```diff
@@ -0,0 +1,73 @@
...
+            except:
+                failed.append(room.data['vidyo_id'])
+                VidyoPlugin.logger.exception('Could not delete VC room %s', room)
+            else:
+                counter += 1
+                room.status = VCRoomStatus.deleted
+                success.append(room.data['vidyo_id'])
+                VidyoPlugin.logger.info('{} deleted', room)
+
+        return {'success': '{}'.format(success),
```
ThiefMaster: why not just return the lists of ids? a repr of a python list is not a very useful return value

vc_vidyo/indico_vc_vidyo/http_api.py
gasparuben: ok

vc_vidyo/indico_vc_vidyo/http_api.py
gasparuben: ok

vc_vidyo/indico_vc_vidyo/http_api.py
gasparuben: It's a kind of control, to avoid massive deletion "by mistake". It's indeed arbitrary number and of course it doesnt preclude the user to invoke the endpoint as many times as wished. Should I remove it?

vc_vidyo/indico_vc_vidyo/http_api.py
gasparuben: ok

vc_vidyo/indico_vc_vidyo/http_api.py
gasparuben: ok

vc_vidyo/indico_vc_vidyo/http_api.py
gasparuben: ok

vc_vidyo/indico_vc_vidyo/http_api.py
gasparuben: ok

vc_vidyo/indico_vc_vidyo/http_api.py
gasparuben: ok

vc_vidyo/indico_vc_vidyo/http_api.py
gasparuben: What do you mean? The return it should be  out of the for loop

ThiefMaster:
Can you please rebase this to `upstream/master`. Somehow one of the commits already in master is showing up again in your PR

vc_vidyo/indico_vc_vidyo/http_api.py
ThiefMaster: `return {'success': success, ...}` since the lists are more useful than those strings you create with the `format()`

vc_vidyo/indico_vc_vidyo/http_api.py
pferreir: Seems a bit pointless to me for the reasons you guys mention. We're specifying each and every ID anyway. And this is only a danger if we have a client that asks for > 10 rooms to be deleted at once. And if such client knows how this works, it will anyway batch the queries.

gasparuben:
I did on Friday Adrian. Please let me know.

vc_vidyo/indico_vc_vidyo/http_api.py
```diff
@@ -0,0 +1,65 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from flask import request
+
+from indico.web.http_api.hooks.base import HTTPAPIHook
+from indico.modules.vc.models.vc_rooms import VCRoom, VCRoomStatus
```
ThiefMaster: modules < web

PR #41, opened 2017-04-19
VC/Vidyo: use zeep instead of suds

pferreir:
It seems like suds' handling of Vidyo WSDL is broken.
[zeep](http://docs.python-zeep.org/en/master/) seems to be its rightful successor, kept up-to-date and designed in a more Pythonic way.


ThiefMaster:
Let's bump the version to 0.4.1? Less hassle to update that way.

pferreir:
@ThiefMaster, I've actually set it to 0.5, since it's a significant change. Maybe we should at some point reset the plugin versions to 1.0 (with 2.0?)

ThiefMaster:
Yeah, 1.0 for everything would be a very good idea for the Indico 2.0 release

PR #43, opened 2017-05-05
WIP: Add xrootd/eos storage plugin

ThiefMaster:
This is mainly useful for Indico instances at CERN, but also for anyone else who happens to have an XRootD/EOS storage system and wants to use it in indico.

### DIFF
--- a/storage_xrootd/setup.py
+++ b/storage_xrootd/setup.py
@@ -20,7 +20,7 @@ from setuptools import setup
 
 setup(
     name='indico_storage_xrootd',
-    version='0.1.dev0',
+    version='0.1.0',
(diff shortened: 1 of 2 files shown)


storage_xrootd/setup.py
```diff
@@ -0,0 +1,42 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
+
+from setuptools import setup
+
+setup(
+    name='indico_storage_xrootd',
+    version='0.1.dev0',
```
ThiefMaster: change to `0.1.0` once it's not WIP anymore

PR #49, opened 2017-07-06
Plugin intallation quick guide

xaionaro:


ThiefMaster:
Thanks, but this will be obsolete once we release 2.0 - the plugins will be on PyPI then. We do need to add something to the indico docs though

PR #62, opened 2018-03-26
S3 storage plugin

### LINKED ISSUE #61: Add an S3 storage plugin
**(Brainstorming - comments welcome!)**

This storage plugin (`storage_s3`) would use *boto3* (or any other *well-maintained* AWS/S3 client) to implement a storage backend (`s3:...`) for attachments etc. backed by S3.

Some functionality the plugin should have:

- Configurable bucket name ([restricted charset](https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html),  i.e. something like `^[a-z0-9-]{3,63}$`)
- Dynamic bucket names - so you can have one bucket per year/month/week (day might be overkill). either via placeholder in the bucket name or separate option (similar to `datestamp` in the eos/xrootd backend)
- Maintenance task (celery): applies to all s3 STORAGE_BACKENDS that use dynamic buckets. create next bucket early enough (e.g. one period before it is needed, e.g. a february bucket would be created on 1st of january, a week 2 bucket would be created beginning of week 1, for yearly it might be enough to do it the month before as well (ie never more than one month before it's needed)). The maintenance task needs to know what ACL etc to set on the bucket.
Specifying an ACL would probably be ugly, let's see if we can copy an existing "template-like" bucket for this?
- S3 credentials should be loadable from a file (maybe using the default format uses for aws credentials) in addition to putting them directly in the config string. we could go one step further and add general support for loading settings from a config file, e.g. something like `s3:config=/path/to/config.file,rotate=weekly` to keep the indico config cleaner when using multiple s3 backends with almost identical configs
- Backup is out of the scope of this plugin; that would be done separately if needed
- Downloading files (for users) is done by redirecting to a pre-signed URL; by default we might want to use the 'virtual' `addressing_style`, but being able to override this probably makes sense. To discuss: always specify content-disposition/content-type with what we have in our database or just specify it during save and assume it remains correct? TTL of the signed links?

While we will most likely use [Ceph's S3 API](http://docs.ceph.com/docs/master/radosgw/s3/) at CERN - also for testing/development - we should make sure the backend works fine with the "real' Amazon S3 as well. The free trial account might do the job, since it includes S3 storage as well.

Useful docs link: http://boto3.readthedocs.io/en/latest/guide/s3.html

meluru:
closes #61


storage_s3/setup.py
```diff
@@ -0,0 +1,44 @@
...
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
+
+from setuptools import setup
+
+
+setup(
+    name='indico-plugin-storage-s3',
+    version='1.0',
```
ThiefMaster: I'd start with `1.0-dev`

storage_s3/setup.py
```diff
@@ -0,0 +1,44 @@
...
+    url='https://github.com/indico/indico-plugins',
+    license='https://www.gnu.org/licenses/gpl-3.0.txt',
+    author='Indico Team',
+    author_email='indico-team@cern.ch',
+    py_modules=('indico_storage_s3',),
+    zip_safe=False,
+    platforms='any',
+    install_requires=[
+        'indico>=2.0',
+        'boto3>=1.6.5'
```
ThiefMaster: Latest version is `1.6.16`, and I'd require a specific version or - if they use semver - `boto3>=1.6.16,<2.0` or `boto3>=1.6.16,<1.7` (depending on what kind of changes they do in 1.x updates)

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -0,0 +1,100 @@
...
+    def get_local_path(self, file_id):
+        with NamedTemporaryFile(suffix='indico.tmp', dir=config.TEMP_DIR) as tmpfile:
+            self._copy_file(self.open(file_id), tmpfile)
+            tmpfile.flush()
+            yield tmpfile.name
+
+    def save(self, name, content_type, filename, fileobj):
+        try:
+            self.client.upload_fileobj(fileobj, self.bucket, name)
+            return name, ''
```
ThiefMaster: The md5 checksum shouldn't be empty

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def getsize(self, file_id):
+        try:
+            return self.client.head_object(Bucket=self.bucket, Key=file_id)['ContentLength']
+        except Exception as e:
+            raise StorageError('Could not get size of "{}": {}'.format(file_id, e)), None, sys.exc_info()[2]
+
+    def send_file(self, file_id, content_type, filename, inline=True):
+        try:
+            fileobj = self.client.get_object(Bucket=self.bucket, Key=file_id)['Body']
```
ThiefMaster: This should generate a presigned URL and redirect to it instead of downloading the file to the server

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -96,7 +96,9 @@ def getsize(self, file_id):
 
     def send_file(self, file_id, content_type, filename, inline=True):
         try:
-            fileobj = self.client.get_object(Bucket=self.bucket, Key=file_id)['Body']
-            return send_file(filename, fileobj, content_type, inline=inline)
+            url = self.client.generate_presigned_url('get_object',
+                                                     Params={'Bucket': self.bucket, 'Key': file_id},
+                                                     ExpiresIn=100)
```
ThiefMaster: 1m40s is a bit strange ;) I'd use 60 or 120.

Also, we may want to include a content-disposition and content-type header here so we always use the one that's specified when `send_file` is called.

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -97,8 +97,11 @@ def getsize(self, file_id):
     def send_file(self, file_id, content_type, filename, inline=True):
         try:
             url = self.client.generate_presigned_url('get_object',
-                                                     Params={'Bucket': self.bucket, 'Key': file_id},
-                                                     ExpiresIn=100)
+                                                     Params={'Bucket': self.bucket,
+                                                             'Key': file_id,
+                                                             'ResponseContentDisposition': 'attachment',
```
ThiefMaster: filename missing ;) also, inline/attachment should depend upon the `inline` argument of this method

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -96,9 +96,15 @@ def getsize(self, file_id):
 
     def send_file(self, file_id, content_type, filename, inline=True):
         try:
+
+            content_disp = ('inline; filename="%s"' % filename if inline
```
ThiefMaster: I would use werkzeug's Headers util to generate this string. While all our filenames should be sanitized to plain ascii without weird chars, it's better to have something that escapes such chars if necessary:

```
>>> from werkzeug.datastructures import Headers
>>> h = Headers(); h.add('Content-Disposition', 'inline', filename=u'x"y.png'); print h['Content-Disposition']
inline; filename="x\"y.png"
```

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def getsize(self, file_id):
+        try:
+            return self.client.head_object(Bucket=self.bucket, Key=file_id)['ContentLength']
+        except Exception as e:
+            raise StorageError('Could not get size of "{}": {}'.format(file_id, e)), None, sys.exc_info()[2]
+
+    def send_file(self, file_id, content_type, filename, inline=True):
+        try:
+
```
mic4ael: blank space?

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -108,3 +111,31 @@ def send_file(self, file_id, content_type, filename, inline=True):
...
+            raise StorageError('Could not create bucket "{}": {}'.format(name, e)), None, sys.exc_info()[2]
+
+    def get_week_of_the_month(self, date):
+        first_day = date.replace(day=1)
+        current_day = date.day
+        return int(ceil((current_day + first_day.weekday())/7))
+
+    def replace_bucket_placeholders(self, name, date):
+        dates = {'year': str(date.year),
+                 'month': str(date.month),
```
ThiefMaster: i'm pretty sure this results in single-digit month numbers ;)

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -108,3 +111,31 @@ def send_file(self, file_id, content_type, filename, inline=True):
...
+                 'week': str(self.get_week_of_the_month(date))}
```
ThiefMaster: I think we want the week number in the year here - `<year>-<month>-<week>` seems strange, IMO `<year>-<week>` is much more sensible

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -108,3 +111,31 @@ def send_file(self, file_id, content_type, filename, inline=True):
...
+        pattern = re.compile("|".join(dates.keys()))
```
ThiefMaster: no need for `.keys()` and single quotes ;)

ThiefMaster: Why not just `re.sub('|'.join(dates), lambda m: ..., name)`? A separate compile step isn't very useful if you only use the compiled regex once.

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -108,3 +111,31 @@ def send_file(self, file_id, content_type, filename, inline=True):
...
+        return pattern.sub(lambda m: dates[re.escape(m.group(0))], name)
```
ThiefMaster: TBH, using a regex here seems strange. Why not simple `replace` calls?

```
name = name.replace('<year>', date.strftime('%Y'))
name = name.replace('<month>', date.strftime('%m'))
name = name.replace('<week>', date.strftime('%W'))  # or %U?
```

could also loop over a dict and call replace in the loop - but I think using a regex seems overkill and reduces readability a lot

storage_s3/indico_storage_s3/__init__.py
```diff
@@ -0,0 +1,8 @@
+from __future__ import unicode_literals
```
ThiefMaster: please add the indico header to non-empty `__init__.py` files

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -0,0 +1,133 @@
...
+    simple_data = False
+
+    def __init__(self, data):
+        data = self._parse_data(data)
+        self.host = data['host']
+        self.bucket = self.get_bucket_name(data['bucket'])
+        self.secret_key = data['secret_key']
+        self.access_key = data['access_key']
+        if any(x in data['bucket'] for x in ['<year>', '<month>', '<week>']):
+            self.acl_template_bucket = data['acl_template_bucket']
```
ThiefMaster: shouldn't this be optional? maybe not everyone wants to set a custom ACL on their bucket.

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -0,0 +1,133 @@
...
+                                                     Params={'Bucket': self.bucket,
+                                                             'Key': file_id,
+                                                             'ResponseContentDisposition': h.get('Content-Disposition'),
+                                                             'ResponseContentType': content_type},
+                                                     ExpiresIn=120)
+            return redirect(url)
+        except Exception as e:
+            raise StorageError('Could not send file "{}": {}'.format(file_id, e)), None, sys.exc_info()[2]
+
+    def get_bucket_name(self, name, replace_placeholders=True):
```
ThiefMaster: I would add a `_` at the beginning of each function that is not part of the normal Storage API

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -0,0 +1,133 @@
...
+        if replace_placeholders:
+            return self.replace_bucket_placeholders(name, datetime.now())
+        else:
+            return name
+
+    def create_bucket(self, name):
+            response = self.client.get_bucket_acl(Bucket=self.acl_template_bucket)
```
ThiefMaster: indentation

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -0,0 +1,133 @@
...
+        except Exception as e:
+            raise StorageError('Could not send file "{}": {}'.format(file_id, e)), None, sys.exc_info()[2]
+
+    def get_bucket_name(self, name, replace_placeholders=True):
+        if replace_placeholders:
+            return self.replace_bucket_placeholders(name, datetime.now())
+        else:
+            return name
+
+    def create_bucket(self, name):
```
ThiefMaster: can we add some logging for this?

storage_s3/indico_storage_s3/task.py
```diff
@@ -0,0 +1,40 @@
...
+from celery.schedules import crontab
+from dateutil.relativedelta import relativedelta
+
+from indico.core.celery import celery
+from indico.core.storage.backend import get_storage
+
+
+@celery.periodic_task(run_every=crontab(minute=0, hour=1))
+def create_bucket():
+    storage = get_storage('s3')
```
ThiefMaster: This assumes the key of the entry in the storage dict is `s3` but that key can be completely arbitrary. You'll have to iterate over all the values in `config.STORAGE_BACKENDS`, check if it's a S3 storage backend (`isinstance`) and then run the logic for it.

storage_s3/indico_storage_s3/task.py
```diff
@@ -0,0 +1,40 @@
...
+    bucket_name = storage.get_bucket_name(replace_placeholders=False)
+    now = datetime.now()
+    if all(x in bucket_name for x in ['<year>', '<week>']) and now.weekday() == 0:
```
ThiefMaster: Isn't a simple `'<year>' in bucket_name and '<week>' in bucket_name` more readable? ;)

storage_s3/indico_storage_s3/task.py
```diff
@@ -0,0 +1,40 @@
...
+from datetime import datetime
+
+from celery.schedules import crontab
+from dateutil.relativedelta import relativedelta
+
+from indico.core.celery import celery
+from indico.core.storage.backend import get_storage
+
+
+@celery.periodic_task(run_every=crontab(minute=0, hour=1))
```
ThiefMaster: once a day should be enough

storage_s3/indico_storage_s3/task.py
```diff
@@ -0,0 +1,40 @@
...
+def create_bucket():
+    storage = get_storage('s3')
+    bucket_name = storage.get_bucket_name(replace_placeholders=False)
+    now = datetime.now()
+    if all(x in bucket_name for x in ['<year>', '<week>']) and now.weekday() == 0:
+        date = now + relativedelta(weeks=1)
+        storage.create_bucket(storage.replace_bucket_placeholders(bucket_name, date))
```
ThiefMaster: I think this will create the bucket more than once since the task will run more than once. Can we check if it already exists?

ThiefMaster: also, let's do some logging in here as well

storage_s3/indico_storage_s3/task.py
```diff
@@ -0,0 +1,40 @@
...
+    elif ((all(x in bucket_name for x in ['<year>', '<month>']) and now.day == 1)
```
ThiefMaster: that three-line if statement isn't very pretty. what about adding a helper function where it's separated into multiple statements and you return whenever you can.

storage_s3/indico_storage_s3/task.py
ThiefMaster: Actually, looking at the checks below, what about extracting the set of placeholders (`re.findall` is probably the easiest option) and using set operations?

    if placeholders & {'<year>', '<month>'}:

This will be especially useful in the checks below

storage_s3/indico_storage_s3/task.py
```diff
@@ -0,0 +1,40 @@
...
+          or ('<year>' in bucket_name and not any(x in bucket_name for x in ['<month>', '<week>'])
```
ThiefMaster: when using a set this whole line can be replaced with `placeholders == {'year'}`

storage_s3/indico_storage_s3/task.py
ThiefMaster: ^ might not apply anymore when using sets; let's see how it looks like after switching to sets?

storage_s3/indico_storage_s3/task.py
meluru: afaik this crontab is once a day at 1 am

ThiefMaster: indeed, i somehow read this as hourly.. not sure why.

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -52,11 +53,10 @@ class S3Storage(Storage):
     def __init__(self, data):
         data = self._parse_data(data)
         self.host = data['host']
-        self.bucket = self.get_bucket_name(data['bucket'])
+        self.bucket = data['bucket']
         self.secret_key = data['secret_key']
         self.access_key = data['access_key']
-        if any(x in data['bucket'] for x in ['<year>', '<month>', '<week>']):
-            self.acl_template_bucket = data['acl_template_bucket']
+        self.acl_template_bucket = data['acl_template_bucket'] if 'acl_template_bucket' in data else None
```
ThiefMaster: self.acl_template_bucket = data.get('acl_template_bucket')

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -0,0 +1,148 @@
...
+            self._copy_file(self.open(file_id), tmpfile)
+            tmpfile.flush()
+            yield tmpfile.name
+
+    def save(self, name, content_type, filename, fileobj):
+        try:
+            fileobject = self._ensure_fileobj(fileobj)
+            bucket = self._get_current_bucket()
+            self.client.upload_fileobj(fileobject, bucket, name)
+            checksum = self.client.head_object(Bucket=bucket, Key=name)['ETag'][1:-1]
```
ThiefMaster: Do we know for sure that this is an md5 checksum (both ceph and s3)? Also out of curiosity, what are the first/last chars you are stripping off?

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -0,0 +1,148 @@
...
+                                                             'ResponseContentDisposition': h.get('Content-Disposition'),
+                                                             'ResponseContentType': content_type},
+                                                     ExpiresIn=120)
+            return redirect(url)
+        except Exception as e:
+            raise StorageError('Could not send file "{}": {}'.format(file_id, e)), None, sys.exc_info()[2]
+
+    def _get_current_bucket(self):
+        if any(x in self.bucket for x in ['<year>', '<month>', '<week>']):
+            return self._replace_bucket_placeholders(self.bucket, datetime.today())
```
ThiefMaster: `date.today()`?

storage_s3/indico_storage_s3/task.py
```diff
@@ -0,0 +1,48 @@
...
+            S3StoragePlugin.logger.info('S3 storage instance found: %s' % key)
+            storage = get_storage(key)
+            now = datetime.now()
+            bucket = storage._get_bucket_name()
+            placeholders = set(re.findall('<.*?>', bucket))
+            if placeholders == {'<year>', '<week>'}:
+                date = now + relativedelta(weeks=1)
+                bucket_name = storage._replace_bucket_placeholders(bucket, date)
+                storage._create_bucket(bucket_name)
+            elif placeholders == {'<year>', '<month>'} or placeholders == {'<year>'} and now.month == 12:
```
ThiefMaster: don't mix `and` and `or` without parentheses

storage_s3/indico_storage_s3/task.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+@celery.periodic_task(run_every=crontab(minute=0, hour=1))
+def create_bucket():
+    storages = get_storage_backends()
+    for key, storage_class in storages.iteritems():
+        if storage_class == S3Storage:
+            S3StoragePlugin.logger.info('S3 storage instance found: %s' % key)
+            storage = get_storage(key)
+            now = datetime.now()
```
ThiefMaster: `date.today()`?

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -0,0 +1,148 @@
...
+                                                             'Key': file_id,
+                                                             'ResponseContentDisposition': h.get('Content-Disposition'),
+                                                             'ResponseContentType': content_type},
+                                                     ExpiresIn=120)
+            return redirect(url)
+        except Exception as e:
+            raise StorageError('Could not send file "{}": {}'.format(file_id, e)), None, sys.exc_info()[2]
+
+    def _get_current_bucket(self):
+        if any(x in self.bucket for x in ['<year>', '<month>', '<week>']):
```
ThiefMaster: IS the check actually needed? `_replace_bucket_placeholders` doesn't do anything if there are no placeholders anyway.

storage_s3/indico_storage_s3/task.py
```diff
@@ -0,0 +1,48 @@
...
+
+from plugin import S3Storage, S3StoragePlugin
+
+
+@celery.periodic_task(run_every=crontab(minute=0, hour=1))
+def create_bucket():
+    storages = get_storage_backends()
+    for key, storage_class in storages.iteritems():
+        if storage_class == S3Storage:
+            S3StoragePlugin.logger.info('S3 storage instance found: %s' % key)
```
ThiefMaster: don't interpolate values into logged strings; that doesn't work well with sentry (which groups by log string). instead, pass them as posargs:

    S3StoragePlugin.logger.info('S3 storage instance found: %s', key)

ThiefMaster: Also, this particular message should be `debug` if you want to keep it. It's just noise and nothing one would need/want in a non-debug log.

storage_s3/indico_storage_s3/task.py
```diff
@@ -0,0 +1,48 @@
...
+from indico.core.celery import celery
+from indico.core.storage.backend import get_storage, get_storage_backends
+
+from plugin import S3Storage, S3StoragePlugin
+
+
+@celery.periodic_task(run_every=crontab(minute=0, hour=1))
+def create_bucket():
+    storages = get_storage_backends()
+    for key, storage_class in storages.iteritems():
```
ThiefMaster: This will iterate over the registered backends - that's not what you need here (it won't work if you have multiple s3 backend instances), and the `get_storage()` below will actually fail if your backend is not named `s3`. Iterate over `config.STORAGE_BACKENDS` instead, then use `isinstance` to check if it's a `S3Storage`.

storage_s3/indico_storage_s3/test.py
```diff
@@ -0,0 +1,71 @@
...
+from freezegun import freeze_time
+
+import indico_storage_s3.plugin as plugin
+from indico_storage_s3.task import create_bucket
+
+
+@pytest.mark.parametrize(('date', 'name_template', 'expected_name'), (
+    ('2018-04-11', 'name', 'name'),
+    ('2018-04-11', 'name-<year>', 'name-2018'),
+    ('2018-04-11', 'name-<year>-<month>', 'name-2018-04'),
```
ThiefMaster: just thinking, should we reject nonsense like month/week without a year?

storage_s3/indico_storage_s3/test.py
```diff
@@ -0,0 +1,71 @@
...
+@mock.patch.object(plugin.S3Storage, '__init__', lambda self: None)
+def test_resolve_bucket_name(date, name_template, expected_name):
+    with freeze_time(date):
+        storage = plugin.S3Storage()
+        storage.bucket = name_template
+        assert storage._get_current_bucket() == expected_name
+
+
+@mock.create_autospec
+def fake_bucket_created(name, replace_placeholders=True):
```
ThiefMaster: I think the term `mock` is more common than `fake` here

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -0,0 +1,148 @@
...
+    name = 's3'
+    simple_data = False
+
+    def __init__(self, data):
+        data = self._parse_data(data)
+        self.host = data['host']
+        self.bucket = data['bucket']
+        self.secret_key = data['secret_key']
+        self.access_key = data['access_key']
+        self.acl_template_bucket = data['acl_template_bucket'] if 'acl_template_bucket' in data else None
```
ThiefMaster: self.acl_template_bucket = data.get('acl_template_bucket')

storage_s3/indico_storage_s3/plugin.py
meluru: What's wrong with `datetime.today()` ?

ThiefMaster: Do you need time information? If yes, it's fine (but I'd use `now()` then which does the same). If not, it seems clearer to just get a date object and not a datetime...

storage_s3/indico_storage_s3/test.py
meluru: I can add test for that, I think it makes no sense to allow it (and the task currently won't create anything in such cases)

ThiefMaster: I meant throwing an exception when instantiating a storage with such a template

storage_s3/indico_storage_s3/plugin.py
meluru: thanks, makes sense :)

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -81,8 +82,9 @@ def save(self, name, content_type, filename, fileobj):
         try:
             fileobject = self._ensure_fileobj(fileobj)
             bucket = self._get_current_bucket()
-            self.client.upload_fileobj(fileobject, bucket, name)
-            checksum = self.client.head_object(Bucket=bucket, Key=name)['ETag'][1:-1]
+            checksum = hashlib.md5(fileobject.read()).hexdigest()
```
ThiefMaster: Please don't read a possibly large file into memory at once ;)

storage_s3/setup.py
```diff
@@ -0,0 +1,44 @@
...
+
+setup(
+    name='indico-plugin-storage-s3',
+    version='1.0-dev',
+    description='S3 storage backend for Indico',
+    url='https://github.com/indico/indico-plugins',
+    license='https://www.gnu.org/licenses/gpl-3.0.txt',
+    author='Indico Team',
+    author_email='indico-team@cern.ch',
+    py_modules=('indico_storage_s3',),
```
ThiefMaster: `packages=find_packages()`

ThiefMaster: remove this line as well, it's not needed

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -0,0 +1,168 @@
...
+                                                     ExpiresIn=120)
+            return redirect(url)
+        except Exception as e:
+            raise StorageError('Could not send file "{}": {}'.format(file_id, e)), None, sys.exc_info()[2]
+
+    def _get_current_bucket(self):
+        return self._replace_bucket_placeholders(self.bucket, date.today())
+
+    def _get_bucket_name(self, current=False):
+            return self._get_current_bucket() if current else self.bucket
```
ThiefMaster: indentation

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def _extend_indico_cli(self, sender, **kwargs):
+        @cli_command()
+        @click.option('--storage', default=None, help='Storage to create bucket for')
+        def create_s3_bucket(storage):
+            """Create s3 bucket"""
+            storages = [storage] if storage else config.STORAGE_BACKENDS
+            for key in storages:
+                storage = get_storage(key)
+                if isinstance(storage, S3Storage):
```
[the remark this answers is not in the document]
ThiefMaster: I don't really see much benefit in that. If you don't have any s3 backends there's nothing to do.

However, what I would indeed change is failing with an error if you specify a storage and that one doesn't exist!

storage_s3/indico_storage_s3/plugin.py
```diff
@@ -0,0 +1,172 @@
...
+    def open(self, file_id):
+        try:
+            s3_object = self.client.get_object(Bucket=self._get_current_bucket(), Key=file_id)['Body']
+            return BytesIO(s3_object.read())
+        except Exception as e:
+            raise StorageError('Could not open "{}": {}'.format(file_id, e)), None, sys.exc_info()[2]
+
+    @contextmanager
+    def get_local_path(self, file_id):
+        with NamedTemporaryFile(suffix='indico.tmp', dir=config.TEMP_DIR) as tmpfile:
```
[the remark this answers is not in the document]
ThiefMaster: Doesn't matter much for a temp file. But I'd go for something like `.s3` - having a suffix not start with a dot/dash is a bit ugly (but then again, it's a temp file so the name isn't really important)

storage_s3/indico_storage_s3/plugin.py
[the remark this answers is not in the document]
meluru: If you specify a storage and that one doesn't exist it will fail with a RuntimeError from `get_storage()` method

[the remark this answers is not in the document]
ThiefMaster: That's not very pretty though ;)

storage_s3/setup.py
```diff
@@ -0,0 +1,44 @@
...
+    description='S3 storage backend for Indico',
+    url='https://github.com/indico/indico-plugins',
+    license='https://www.gnu.org/licenses/gpl-3.0.txt',
+    author='Indico Team',
+    author_email='indico-team@cern.ch',
+    packages=find_packages(),
+    zip_safe=False,
+    platforms='any',
+    install_requires=[
+        'indico>=2.1rc2',
```
ThiefMaster: you can change this to rc3 now that we have a rc3 that contains your change

storage_s3/indico_storage_s3/plugin.py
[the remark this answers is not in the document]
meluru: @mic4ael why do you think it would be much better to indicate it's s3 in tmp file name?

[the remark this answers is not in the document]
mic4ael: It seems to be more verbose to me and if by any accident (not sure how probable it is) the file doesn't get removed I would be able to tell which part of the system failed.

ThiefMaster:
Idea: Store some extra metadata in `x-amz-meta-*` headers. Could be the Attachment ID for example. Optional `meta` dict in `Storage.save()`. Would break compatibility with existing storage backends (since their save() methods wouldn't have this argument), but I don't think anyone wrote their own so far.

storage_s3/indico_storage_s3/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
+
+from indico.core import signals
+from indico.util.i18n import make_bound_gettext
+
+
+_ = make_bound_gettext('search_cern')
```
pferreir: Copy-paste issue?

storage_s3/indico_storage_s3/controllers.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def _get_static_info(self, storage):
+        return {
+            'bucket': storage.bucket_name,
+            'dynamic': False,
+        }
+
+    def _get_dynamic_info(self, backend_name, storage):
+        buckets = set()
+        for model in StoredFileMixin.__subclasses__():
```
pferreir: Is this kind of introspection OK in a plugin? A signal returning the same thing from the core would be already better.

storage_s3/indico_storage_s3/controllers.py
```diff
@@ -0,0 +1,77 @@
...
+            query = (db.session.query(db.func.split_part(model.storage_file_id, '//', 1))
+                     .filter(model.storage_file_id.isnot(None), model.storage_backend == backend_name))
```
pferreir: Missing a `DISTINCT`, as discussed on IRC.

storage_s3/indico_storage_s3/controllers.py
ThiefMaster: IMHO yes: I think when using a signal for this people would be likely to forget registering it (since it wouldn't break anything else). Also, this applies to any subclass of `StorageFileMixin`.

PR #66, opened 2018-05-17
Payment Manual: Implemented pending transactions

driehle:
Currently, when the `payment_manual` plugin is used for bank transfers, the registration will be marked as unpaid for quite a long time. This is irritating for the user, who probably has already initiated the bank transfer.

This pull request adds an optional feature to `payment_manual`, which adds a pending transaction to the registration. The feature can be configured on a per-event basis in the management area:

plugin-settings

If enabled, users have to explicitly confirm the checkout by ticking a checkbox:

plugin-checkout

After the explicit checkout, a pending transaction is created in the database. Together with indico/indico#3361 this prevents that users can change anything about their registration, neither can they check out a second time. In the management area, a manager can approve the receival of the transfer:

plugin-transaction

### DIFF
--- a/payment_manual/indico_payment_manual/blueprint.py
+++ b/payment_manual/indico_payment_manual/blueprint.py
@@ -1,0 +1,28 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2018 European Organization for Nuclear Research (CERN).
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 3 of the
+# License, or (at your option) any later version.
+#
+# Indico is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
+
+from indico.core.plugins import IndicoPluginBlueprint
+
+from indico_payment_manual.controllers import RHManualConfirm, RHManualReceived
+
+
+blueprint = IndicoPluginBlueprint('payment_manual', __name__,
+                                  url_prefix='/event/<confId>/registrations/<int:reg_form_id>/payment/manual')
+
+blueprint.add_url_rule('/confirm', 'confirm', RHManualConfirm, methods=('GET',))
+blueprint.add_url_rule('/received', 'received', RHManualReceived, methods=('GET',))
--- a/payment_manual/indico_payment_manual/controllers.py
+++ b/payment_manual/indico_payment_manual/controllers.py
@@ -1,0 +1,74 @@
+# This file is part of Indico.
+# Copyright (C) 2002 - 2018 European Organization for Nuclear Research (CERN).
+#
+# Indico is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 3 of the
+# License, or (at your option) any later version.
+#
+# Indico is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
+
+from flask import flash, redirect, request, session
+from werkzeug.exceptions import BadRequest
+
+from indico.modules.events.payment.models.transactions import TransactionAction
+from indico.modules.events.payment.util import register_transaction
+from indico.modules.events.registration.models.registrations import Registration
+from indico.modules.events.registration.controllers.management import RHManageRegFormsBase
+from indico.web.flask.util import url_for
+from indico.web.flask.templating import get_template_module
+from indico.web.rh import RH
+from indico.web.util import jsonify_data
+
+from indico_payment_paypal import _
+
+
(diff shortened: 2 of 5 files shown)


payment_manual/indico_payment_manual/blueprint.py
```diff
@@ -0,0 +1,28 @@
...
+
+from indico.core.plugins import IndicoPluginBlueprint
+
+from indico_payment_manual.controllers import RHManualConfirm, RHManualReceived
+
+
+blueprint = IndicoPluginBlueprint('payment_manual', __name__,
+                                  url_prefix='/event/<confId>/registrations/<int:reg_form_id>/payment/manual')
+
+blueprint.add_url_rule('/confirm', 'confirm', RHManualConfirm, methods=('GET',))
```
ThiefMaster: GET is the default, no need to specify `methods` at all in this case.

payment_manual/indico_payment_manual/controllers.py
```diff
@@ -0,0 +1,74 @@
...
+from indico.modules.events.registration.controllers.management import RHManageRegFormsBase
+from indico.web.flask.util import url_for
+from indico.web.flask.templating import get_template_module
+from indico.web.rh import RH
+from indico.web.util import jsonify_data
+
+from indico_payment_paypal import _
+
+
+class RHManualConfirm(RH):
```
ThiefMaster: Inherit from `RHRegistrationFormRegistrationBase` and you don't need the custom `_process_args` method.

payment_manual/indico_payment_manual/templates/transaction_details.html
```diff
@@ -0,0 +1,39 @@
...
+{% from 'message_box.html' import message_box %}
+
+{% block details %}
+    {% if transaction.status.name == 'pending' %}
+        <dt>{% trans %}Status{% endtrans %}</dt>
+        <dd>{% trans %}Awaiting bank transfer{% endtrans %}</dd>
+        <dt>&nbsp;</dt>
+        <dd>
+            <a class="i-button"
+               href="{{ url_for('plugin_payment_manual.received', transaction.registration.locator.uuid) }}">
```
ThiefMaster: With the change to the RH inheritance suggested above, use `transaction.registration`. That will use the registration ID which is more appropriate in the management area.

payment_manual/indico_payment_manual/controllers.py
```diff
@@ -0,0 +1,74 @@
...
+                             currency=self.registration.currency,
+                             action=TransactionAction.pending,
+                             provider='manual',
+                             data={})
+
+        flash(_('Please wait two to three business days until we have received your payment.'), 'info')
+        return redirect(url_for('event_registration.display_regform', self.registration.locator.registrant))
+
+
+class RHManualReceived(RHManageRegFormsBase):
```
ThiefMaster: This looks like a management action => inherit from `RHManageRegistrationBase` and get rid of the `_process_args` method.

payment_manual/indico_payment_manual/plugin.py
```diff
@@ -26,30 +27,37 @@
...
 
 class PluginSettingsForm(PaymentPluginSettingsFormBase):
     details = TextAreaField(_('Payment details'), [])
+    trigger_pending = BooleanField(_('Trigger pending state'), [])
 
     def __init__(self, *args, **kwargs):
         super(PluginSettingsForm, self).__init__(*args, **kwargs)
         self.details.description = DETAILS_DESC + '\n' + render_placeholder_info('manual-payment-details',
                                                                                  regform=None, registration=None)
+        self.trigger_pending.description = TRIGGER_DESC
```
ThiefMaster: Set this directly on the field (like it's done in most other forms). The only reason why we handle the description in here is because we add a dynamic part.

payment_manual/indico_payment_manual/templates/event_payment_form.html
```diff
@@ -1 +1,23 @@
 {{ details | markdown }}
+
+{% if trigger_pending %}
+    <div class="checkbox-with-text">
+        <input type="checkbox" id="payment-manual-agree">
+        <div>
+            {% trans %}I understand that the registration will be completed only after the bank transfer is received. A bank trasfer usually takes two to three business days. Until then the status of your registration will remain „awaiting payment“. You will receive an email confirming your payment once the bank transfer is completed.{% endtrans %}
```
ThiefMaster: that line is way too long

payment_manual/indico_payment_manual/templates/event_payment_form.html
```diff
@@ -1 +1,23 @@
...
+        </div>
+    </div>
+    <p class="text-right">
+        <button type="button" id="payment-manual-confirm" class="action-button" data-url="{{ confirm_url }}">{% trans %}Confirm{% endtrans %}</button>
+    </p>
+    <script type="text/javascript">
+        $("#payment-manual-confirm").on("click", function(e) {
+            if ($("#payment-manual-agree").is(":checked")) {
```
ThiefMaster: `.prop('checked')`

payment_manual/indico_payment_manual/templates/transaction_details.html
```diff
@@ -0,0 +1,39 @@
...
+{% endblock %}
+
+{% block warning_box %}
+    {% if transaction.amount != transaction.registration.price %}
+        {% call message_box('warning') %}
+            <p>
+                {%- trans %}The paid amount does not match the required amount. Please contact the registrant to solve the issue.{% endtrans -%}
+            </p>
+            <p>
+                {%- trans %}Paid: {% endtrans -%}
```
ThiefMaster: Don't include the trailing space here. Remove the `-` from `{{-` in the next line instead so Jinja doesn't collapse the space.

payment_manual/indico_payment_manual/templates/transaction_details.html
```diff
@@ -0,0 +1,39 @@
...
+                {{- format_currency(transaction.amount, transaction.currency, locale=session.lang) -}}
+            </p>
+            <p>
+                {%- trans %}Required: {% endtrans -%}
```
ThiefMaster: same

PR #67, opened 2018-08-13
Add ursh plugin for indico

errikos:
- [x] Add a new blueprint path: `/ursh`
- [x] Add a template hook (`ursh`) to enable short link generation
- [x] Add a web page `/url-shortener`, where users can shorten an arbitrary Indico URL
- [x] Add capability to create custom shortcuts


ursh/indico_ursh/plugin.py
```diff
@@ -0,0 +1,67 @@
...
+from indico_ursh import _
+from indico_ursh.blueprint import blueprint
+
+
+class SettingsForm(IndicoForm):
+    api_key = StringField(_('API key'), [DataRequired()],
+                          description=_('The API key to access the <tt>ursh</tt> service'))
+    server = URLField(_('Shortening server'), [DataRequired()],
+                      description=_('The <tt>ursh</tt> shortening server (which generates short URLs)'))
+    host = URLField(_('Redirection host'), [DataRequired()],
```
ThiefMaster: We could make this optional and use the `server` if it's not set

ursh/indico_ursh/plugin.py
```diff
@@ -0,0 +1,67 @@
...
+
+class UrshPlugin(IndicoPlugin):
+    """URL Shortener
+
+    Provides a URL shortening service for indico assets, events, etc.
+    """
+
+    configurable = True
+    settings_form = SettingsForm
+    category = PluginCategory.other
```
ThiefMaster: that's implied

ursh/setup.py
```diff
@@ -0,0 +1,43 @@
...
+    description='URL shortening service for Indico',
+    url='https://github.com/indico/indico-plugins',
+    license='https://www.gnu.org/licenses/gpl-3.0.txt',
+    author='Indico Team',
+    author_email='indico-team@cern.ch',
+    packages=find_packages(),
+    zip_safe=False,
+    include_package_data=True,
+    install_requires=[
+        'indico>=2.0',
```
ThiefMaster: `>=2.2.dev0` since it'll require new template hooks that won't be there before 2.2

ursh/indico_ursh/plugin.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def init(self):
+        super(UrshPlugin, self).init()
+        self.template_hook('ursh', self._inject_ursh_button)
+        self.inject_bundle('main.js', WPBase)
+
+    @property
```
ThiefMaster: doesn't need to be a property; setting it directly on the class is fine

ursh/indico_ursh/plugin.py
```diff
@@ -0,0 +1,67 @@
...
+    def default_settings(self):
+        return {
+            'api_key': '',
+            'host': '',
+            'server': '',
+        }
+
+    def get_blueprints(self):
+        yield blueprint
```
ThiefMaster: you can `return` it since there's only one

ursh/indico_ursh/plugin.py
```diff
@@ -0,0 +1,67 @@
...
+    Provides a URL shortening service for indico assets, events, etc.
+    """
+
+    configurable = True
+    settings_form = SettingsForm
+    category = PluginCategory.other
+
+    def init(self):
+        super(UrshPlugin, self).init()
+        self.template_hook('ursh', self._inject_ursh_button)
```
ThiefMaster: this hook name is way too specific. let's find a name that suits the position where you put it (in that case you'd probably have quite a few differently named hooks) or use something like `url-shortener-link`. then it'd be clear what the hook is meant for, even when using it on completely different pages

ursh/indico_ursh/plugin.py
```diff
@@ -0,0 +1,67 @@
...
+from indico.web.views import WPBase
+
+from indico_ursh import _
+from indico_ursh.blueprint import blueprint
+
+
+class SettingsForm(IndicoForm):
+    api_key = StringField(_('API key'), [DataRequired()],
+                          description=_('The API key to access the <tt>ursh</tt> service'))
+    server = URLField(_('Shortening server'), [DataRequired()],
```
ThiefMaster: `api_host`?

ThiefMaster:
an empty `MANIFEST.in` isn't very useful. since your plugin contains js and has translatable content, put this in the manifest:

```
graft indico_ursh/static
graft indico_ursh/translations

global-exclude *.pyc __pycache__ .keep
```

ursh/indico_ursh/plugin.py
errikos: Maybe, but in the case of `ursh` this does not make sense - except if we use the host of the `server`.

ThiefMaster: It is possible to have API and redirector on the same host

errikos: What I mean is that API will probably be under another blueprint (`/urls/` in our case). So is it safe to just drop the path and only use the host for redirection?

ThiefMaster: Thinking about this again, why do we need the host setting at all? I think ursh should return an absolute URL when creating a new shorturl!

errikos: As it is, ursh returns only the shortcut. Makes sense to also return its `BASE_URL`.

ursh/indico_ursh/controllers.py
```diff
@@ -41,25 +42,33 @@ def _resolve_full_url(original_url):
     @staticmethod
     def _request_short_url(original_url):
         from indico_ursh.plugin import UrshPlugin
+        api_key = UrshPlugin.settings.get('api_key')
+        api_host = UrshPlugin.settings.get('api_host')
+        if not api_key or not api_host:
+            raise BadRequest(_('API key/host missing - have you set the plugin settings?'))
```
ThiefMaster: we do not translate errors that are not commonly shown for users

ursh/indico_ursh/controllers.py
```diff
@@ -41,25 +42,33 @@ def _resolve_full_url(original_url):
     @staticmethod
```
ThiefMaster: I would make this simply a standalone function in `util.py` instead of having it a static method here

ursh/indico_ursh/controllers.py
```diff
@@ -41,25 +42,33 @@ def _resolve_full_url(original_url):
...
         headers = {'Authorization': 'Bearer {api_key}'.format(api_key=UrshPlugin.settings.get('api_key'))}
         response = requests.post(UrshPlugin.settings.get('api_host'),
                                  data=dict(url=original_url, allow_reuse=True),
                                  headers=headers).json()
         return posixpath.join(response['host'], response['shortcut'])
 
+    @staticmethod
+    def _check_host(full_url):
+        if url_parse(full_url).host != url_parse(config.BASE_URL).host:
+            raise BadRequest(_('Invalid host for URL shortening service'))
```
ThiefMaster: no need to translate; users only see this when tinkering with the requests

ursh/indico_ursh/controllers.py
```diff
@@ -41,25 +42,33 @@ def _resolve_full_url(original_url):
...
-            if url_parse(full_url).host != url_parse(config.BASE_URL).host:
-                raise BadRequest('Invalid host for URL shortening service')
+            self._check_host(full_url)
             short_url = self._request_short_url(full_url)
             return jsonify(success=True, msg=short_url)
         except BadRequest as e:
             return jsonify(success=False, msg=e.description)
         except RequestException:
-            return jsonify(success=False, msg='Failed to connect to URL shortening service')
+            return jsonify(success=False, msg=_('Failed to connect to URL shortening service'))
```
ThiefMaster: I would just let it fail - that way users get the error report dialog and can send a report if they care enough. Also, the error will go to sentry in that case (if enabled)

ursh/indico_ursh/controllers.py
```diff
@@ -41,25 +42,33 @@ def _resolve_full_url(original_url):
...
         except (KeyError, ValueError):
```
ThiefMaster: do you expect this to happen? if not, don't catch it

ursh/indico_ursh/controllers.py
```diff
@@ -41,25 +42,33 @@ def _resolve_full_url(original_url):
...
         return posixpath.join(response['host'], response['shortcut'])
 
+    @staticmethod
+    def _check_host(full_url):
+        if url_parse(full_url).host != url_parse(config.BASE_URL).host:
+            raise BadRequest(_('Invalid host for URL shortening service'))
+
     def _process(self):
         try:
             original_url = request.json.get(u'original_url')
```
ThiefMaster: we use the `unicode_literal` futures in all files; no need for the `u` prefix

ursh/indico_ursh/controllers.py
```diff
@@ -41,25 +42,33 @@ def _resolve_full_url(original_url):
...
             full_url = self._resolve_full_url(original_url)
-            if url_parse(full_url).host != url_parse(config.BASE_URL).host:
-                raise BadRequest('Invalid host for URL shortening service')
+            self._check_host(full_url)
             short_url = self._request_short_url(full_url)
             return jsonify(success=True, msg=short_url)
         except BadRequest as e:
```
ThiefMaster: no need to catch this one either; it won't hapen during normal usage and the standard error handling is fine otherwise

ursh/indico_ursh/controllers.py
```diff
@@ -20,6 +20,7 @@
 
 import requests
 from flask import jsonify, request
+from indico_ursh import _
```
ThiefMaster: import order; imports from the current package go in the last block of imports (ie a new one below the `indico.*` imports)

ursh/indico_ursh/util.py
```diff
@@ -0,0 +1,34 @@
...
+
+import requests
+
+
+def request_short_url(original_url):
+    from indico_ursh.plugin import UrshPlugin
+    api_key = UrshPlugin.settings.get('api_key')
+    api_host = UrshPlugin.settings.get('api_host')
+    if not api_key or not api_host:
+        raise BadRequest('API key/host missing - have you set the plugin settings?')
```
ThiefMaster: I'd raise a `ServiceUnavailable` error here; it's not related to the client sending the request doing something wrong. I'd also remove the part after the dash, since it's not useful to non-admins. Actually, I would change the plugin signal handlers so they only add the url shortener link if key and host are set.

ursh/indico_ursh/util.py
```diff
@@ -0,0 +1,34 @@
...
+    headers = {'Authorization': 'Bearer {api_key}'.format(api_key=UrshPlugin.settings.get('api_key'))}
+    response = requests.post(UrshPlugin.settings.get('api_host'),
+                             data=dict(url=original_url, allow_reuse=True),
+                             headers=headers).json()
```
ThiefMaster: I would do a `response.raise_for_status()` before trying to get its JSON data.

ursh/setup.py
```diff
@@ -0,0 +1,43 @@
...
+# along with Indico; if not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
+
+from setuptools import find_packages, setup
+
+
+setup(
+    name='indico-plugin-ursh',
+    version='1.0',
```
ThiefMaster: `3.0-dev` since it requires indico 2.2, and is currently under development. Basically 1.x plugins are for indico 2.0, 2.x plugins require 2.1, 3.x plugins require 2.2, ...

ursh/indico_ursh/client/index.js
```diff
@@ -18,40 +18,41 @@
...
+async function _makeUrshRequest(originalURL, event) {
+    const urshEndpoint = '/ursh';
+
+    let response;
+    try {
+        response = await indicoAxios.post(urshEndpoint, {
+            original_url: originalURL,
+        });
+    } catch (error) {
+        handleAxiosError(error);
```
ThiefMaster: return after that, otherwise things go kaboom :bomb:

ursh/indico_ursh/client/index.js
```diff
@@ -18,40 +18,41 @@
...
+}
+
 
-    $('.ursh-get').on('click', (event) => {
-        const originalURL = $(event.target).attr('data-original-url');
-        return _makeUrshRequest(originalURL, event);
-    });
+$(document).on('click', '.ursh-get', (event) => {
+    const originalURL = $(event.target).attr('data-original-url');
+    return _makeUrshRequest(originalURL, event);
```
ThiefMaster: no need to return that. but you might want to `event.preventDefault()` to avoid getting a `#` in the url from clicking the link

ursh/indico_ursh/util.py
```diff
@@ -16,19 +16,17 @@
...
-
 import requests
 
 
 def request_short_url(original_url):
     from indico_ursh.plugin import UrshPlugin
     api_key = UrshPlugin.settings.get('api_key')
     api_host = UrshPlugin.settings.get('api_host')
-    if not api_key or not api_host:
-        raise BadRequest('API key/host missing - have you set the plugin settings?')
```
ThiefMaster: I would keep raising an error if the values aren't set (after all, someone could send a request to the endpoint manually). Just not a BadRequest. Something as simple as `ServiceUnavailable('Not configured')` should be fine.

ursh/indico_ursh/plugin.py
```diff
@@ -59,4 +59,4 @@ def get_blueprints(self):
 
     def _inject_ursh_button(self, target, **kwargs):
         if self.settings.get('api_key') and self.settings.get('api_host'):
-            return render_plugin_template('ursh_button.html', target=target)
+            return render_plugin_template('ursh_button.html', target=target, **kwargs)
```
ThiefMaster: `**kwargs` of a signal handler shouldn't be used; it's just for forward compatibility with new arguments added to a signal. Any param you expect to handle should be defined explicitly - this is better here anyway since you can now give it a default value ;)

ursh/indico_ursh/templates/ursh_button.html
```diff
@@ -1,3 +1,5 @@
-<span>
-    <a class="ursh-get" data-original-url="{{ target }}">(short-link)</a>
-</span>
+<{{ element }} class="ursh-get {{ element_class }}"
+               data-original-url="{{ target }}"
```
ThiefMaster: we usually define `data-*` attributes after all non-data attributes

ursh/indico_ursh/client/index.js
```diff
@@ -0,0 +1,60 @@
...
+        });
+    } catch (error) {
+        handleAxiosError(error);
+        return;
+    }
+
+    const data = response.data;
+    if (data.success) {
+        $(triggerElement).copyURLTooltip(data.url).show();
+    } else {
```
ThiefMaster: We don't return `success=False` anymore, so that else block could go away (and so can the `success=True` from the response)

errikos: True, quick-pushed and did not catch that.

ursh/indico_ursh/views.py
```diff
@@ -0,0 +1,28 @@
...
+from __future__ import unicode_literals
+
+from flask_pluginengine import render_plugin_template
+
+from indico.core.plugins import WPJinjaMixinPlugin
+from indico.web.views import WPDecorated
+
+
+class WPShortenURLPage(WPJinjaMixinPlugin, WPDecorated):
+
```
ThiefMaster: no empty line here (the empty line is after a class-level docstring, which we don't have here)

ursh/indico_ursh/views.py
```diff
@@ -0,0 +1,28 @@
...
+    def _getBody(self, params):
+        return render_plugin_template('url_shortener.html', **params)
```
ThiefMaster: check in other places where we render templates from plugins; we usually call `WPSomething.render_template('foo.html')` instead of manually instantiating the WP and calling display (that's kind of legacy)

ursh/indico_ursh/templates/url_shortener.html
```diff
@@ -0,0 +1,25 @@
...
+            {% trans -%}
+                In addition to being able to generate short URLs for predefined assets, e.g. events, contributions,
+                etc., you can also generate short URLs for any desired Indico URL by filling it in the input below and
+                clicking the "Shorten" button.
+            {%- endtrans %}
+        </p>
+        <div class="i-has-action">
+            <input id="ursh-shorten-input" name="ursh-shorten-original-url" type="url" style="font-family:monospace;"
+                   spellcheck="false" autocomplete="off" autocapitalize="off"
+                   placeholder="Enter an Indico URL to shorten">
```
ThiefMaster: `{% trans %}...{% endtrans %}`

ursh/indico_ursh/templates/url_shortener.html
```diff
@@ -0,0 +1,25 @@
...
+
+{% block title -%}
+    {% trans %}URL Shortener{% endtrans %}
+{%- endblock %}
+
+{% block content -%}
+    <div class="page-content">
+        <p>
+            {% trans -%}
+                In addition to being able to generate short URLs for predefined assets, e.g. events, contributions,
```
ThiefMaster: I would just explain this particular page here. "You can generate short urls for any Indico URL here" - something like that. And then maybe as a tip mention that there are also quicklinks in various places to get a shorturl.

ursh/indico_ursh/controllers.py
```diff
@@ -26,26 +26,38 @@
...
+        if url_parse(original_url).host:
             return original_url
+        if require_absolute:
+            raise BadRequest('Invalid URL for URL shortening service')
         original_url = original_url.lstrip('/')
         return posixpath.join(config.BASE_URL, original_url)
 
     @staticmethod
     def _check_host(full_url):
+        print(full_url, url_parse(full_url).host)
```
ThiefMaster: :banana:

ursh/indico_ursh/controllers.py
```diff
@@ -26,26 +26,38 @@
...
         if url_parse(full_url).host != url_parse(config.BASE_URL).host:
             raise BadRequest('Invalid host for URL shortening service')
 
     def _process(self):
         original_url = request.json.get('original_url')
-        full_url = self._resolve_full_url(original_url)
+        require_absolute = request.json.get('absolute')
```
ThiefMaster: why not just check client-side in JS if it's absolute? since the value here is client-provided anyway there's no benefit in doing it on the server

ursh/indico_ursh/client/index.js
```diff
@@ -90,11 +90,18 @@ $(document)
         evt.preventDefault();
         const originalURL = _getUrshInput();
         if (originalURL) {
-            _makeUrshRequest(originalURL, evt.target);
+            _makeUrshRequest(originalURL, (result) => {
```
pferreir: this would instead be:
```js
const result = await _makeUrshRequest(originalURL);
const outputElement = // ...
```

ursh/indico_ursh/client/index.js
```diff
@@ -90,11 +90,18 @@ $(document)
...
+                outputElement.val(result);
+                outputElement.select();
+            });
         }
     })
     .on('click', '.ursh-get', (evt) => {
         evt.preventDefault();
         const originalURL = $(evt.target).attr('data-original-url');
-        _makeUrshRequest(originalURL, evt.target);
+        _makeUrshRequest(originalURL, (result) => {
```
pferreir: Same here.

ursh/indico_ursh/client/index.js
```diff
@@ -32,7 +32,7 @@ async function _makeUrshRequest(originalURL, triggerElement) {
     }
 
     const data = response.data;
-    $(triggerElement).copyURLTooltip(data.url).show();
+    callback(data.url);
```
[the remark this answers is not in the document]
errikos: Hmm, but then shouldn't we wrap the call to an `async` function?

ursh/indico_ursh/client/index.js
```diff
@@ -0,0 +1,122 @@
...
+        return;
+    }
+
+    return response.data.url;
+}
+
+function _validateAndFormatURL(url) {
+    const $t = $T.domain('ursh');
+    if (!url) {
+        throw $t.gettext('Please fill in a URL to shorten');
```
ThiefMaster: I'm strongly against throwing strings; we actually have `no-throw-literal` enabled in our eslint config.

ursh/indico_ursh/client/index.js
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    // if protocol is missing, prepend it
+    if (url.startsWith(window.location.hostname)) {
```
ThiefMaster: `window.` is redundant here

ursh/indico_ursh/client/index.js
```diff
@@ -0,0 +1,122 @@
...
+        url = `${window.location.protocol}//${url}`;
+    }
+
+    let parsedURL;
+    try {
+        parsedURL = new URL(url);
```
ThiefMaster: https://caniuse.com/#feat=url :crying_cat_face:

ursh/indico_ursh/client/index.js
```diff
@@ -0,0 +1,122 @@
...
+    evt.preventDefault();
+    const originalURL = $(evt.target).attr('data-original-url');
+    const result = await _makeUrshRequest(originalURL);
+    $(evt.target).copyURLTooltip(result);
+}
+
+$(document)
+    .on('click', '#ursh-shorten-button', _handleUrshPageInput)
+    .on('keydown', '#ursh-shorten-input', (evt) => {
+        if (evt.key === 'Enter') {
```
ThiefMaster: nice, didn't know about this property

ursh/indico_ursh/client/index.js
```diff
@@ -0,0 +1,122 @@
...
+            $('#ursh-shorten-response-form').slideDown();
+            outputElement.val(result);
+            outputElement.select();
+        }
+    }
+}
+
+async function _handleUrshClick(evt) {
+    evt.preventDefault();
+    const originalURL = $(evt.target).attr('data-original-url');
```
ThiefMaster: I think you could get it directly using `evt.target.dataset.originalUrl`

ThiefMaster: Nothing wrong with using jQuery though; but since you create an object for evt.target twice, I'd do `const $elem = $(evt.target);` and use that in both places.

ursh/indico_ursh/client/index.js
errikos: What is the problem? Too small coverage? IE?

ursh/indico_ursh/client/index.js
errikos: Yeah found it while I was looking for an alternative to `keyCode === 13`.

ursh/indico_ursh/client/index.js
ThiefMaster: IE11.. and I guess a ponyfill for it is quite big?

errikos: I guess I could use regular expressions, but this seemed cleaner and less error-prone...

ursh/indico_ursh/client/index.js
```diff
@@ -46,7 +63,8 @@ function _validateAndFormatURL(url) {
     }
 
     // regular expression, because I.E. does not like the URL class
-    const re = RegExp(`^(.+:)//([^/]+)(/.*)?$`);
+    // provides minimal validation, leaving the serious job to the server
+    const re = RegExp(`^([\\d\\w]+:)//([^/ .]+(?:\\.[^/ .]+)*)(/.*)?$`);
```
ThiefMaster: no need for backticks, you don't interpolate anything. fwiw, i would use `/.../` instead of `new RegExp()` (are you even supposed to use the constructor without `new`?), even if that requires some escaping of forward slashes. that's mostly a matter of preference though

ursh/indico_ursh/client/index.js
```diff
@@ -65,48 +83,37 @@ function _validateAndFormatURL(url) {
...
+        _showTip(input, err.message);
         input.focus().select();
         return null;
     }
 }
 
 async function _handleUrshPageInput(evt) {
     evt.preventDefault();
-    const originalURL = _getUrshInput();
+    const $t = $T.domain('ursh');
```
ThiefMaster: this should be safe to do on the global module scope outside any function

ursh/indico_ursh/plugin.py
```diff
@@ -31,6 +32,9 @@
 from wtforms.validators import DataRequired
 
 
+logger = Logger.get('ursh')
```
ThiefMaster: You can use `UrshPlugin.logger` - no need to create your own

ursh/indico_ursh/templates/ursh_link.html
```diff
@@ -0,0 +1,33 @@
+{% set event_manager = event and event.can_manage(session.user) %}
+{% set target = event.url if event else target %}
+{% if dropdown or event_manager %}
+    <a href="" class="arrow js-dropdown ursh-dropdown {{ element_class }}" data-toggle="dropdown"
+       title="{% trans %}Short URL options{% endtrans %}"></a>
+    <ul class="dropdown">
+        <li>
+            <a class="ursh-get"
+               data-original-url="{{ target }}">
```
ThiefMaster: fits in the previous line

ursh/indico_ursh/templates/ursh_link.html
```diff
@@ -0,0 +1,33 @@
+{% set event_manager = event and event.can_manage(session.user) %}
+{% set target = event.url if event else target %}
+{% if dropdown or event_manager %}
+    <a href="" class="arrow js-dropdown ursh-dropdown {{ element_class }}" data-toggle="dropdown"
```
ThiefMaster: `href="#"` or omit it (if the styling is still fine then)

ursh/indico_ursh/templates/ursh_link.html
```diff
@@ -0,0 +1,33 @@
...
+       title="{% trans %}Short URL options{% endtrans %}"></a>
+    <ul class="dropdown">
+        <li>
+            <a class="ursh-get"
+               data-original-url="{{ target }}">
+                {{ trans }}Obtain short URL{{ end_trans }}
+            </a>
+        </li>
+        {%- if event_manager %}
+        <li>
```
ThiefMaster: missing indentation

errikos: this was on purpose :smirk:

ursh/indico_ursh/client/index.js
```diff
@@ -132,4 +136,22 @@ $(document)
             _handleUrshPageInput(evt);
         }
     })
-    .on('click', '.ursh-get', _handleUrshClick);
+    .on('click', '.ursh-get', _handleUrshClick)
+    .on('keyup blur', '#ursh-custom-shortcut-input', (evt) => {
```
ThiefMaster: `input` event would be more suitable here

ursh/indico_ursh/client/index.js
```diff
@@ -132,4 +136,22 @@ $(document)
...
+        const value = $(evt.target).val();
+        if (!_validateUrshCustomShortcut(value)) {
+            $('#ursh-custom-shortcut-submit-button').attr('disabled', true);
```
ThiefMaster: `.prop('disabled', !_validateUrshCustomShortcut(value))`

ursh/indico_ursh/client/index.js
```diff
@@ -132,4 +136,22 @@ $(document)
...
+        } else {
+            $('#ursh-custom-shortcut-submit-button').attr('disabled', false);
+        }
+    })
+    .on('mouseenter', '#ursh-custom-shortcut-submit-button', (evt) => {
+        if ($(evt.target).attr('disabled') === 'disabled') {
```
ThiefMaster: `.prop('disabled')` gives you a boolean

ThiefMaster: or just `evt.target.disabled`, no need for jquery here

ursh/indico_ursh/controllers.py
```diff
@@ -52,8 +56,52 @@ def _process(self):
...
-class RHDisplayShortenURLPage(RH):
+class RHShortURLPage(RH):
     """Provide a simple page, where users can submit a URL to be shortened"""
 
     def _process(self):
-        return WPShortenURLPage.render_template('url_shortener.html')
+        return WPShortenURLPage.render_template('ursh_shortener_page.html')
+
+
+class RHCustomShortURLPage(RHEventBase):
```
ThiefMaster: this should inherit from the event management RH since only event managers may use it

ursh/indico_ursh/controllers.py
```diff
@@ -52,8 +56,52 @@ def _process(self):
...
+            return _('Shortcut already exists')
+
+    def _process_args(self):
+        super(RHCustomShortURLPage, self)._process_args()
+        from indico_ursh.plugin import UrshPlugin
+        api_host = url_parse(UrshPlugin.settings.get('api_host'))
+        self.ursh_host = strip_end(api_host.to_url(), api_host.path[1:])
+
+    def _process_GET(self):
+        original_url = self._make_absolute_url(request.args.get('original_url'))
```
ThiefMaster: `request.args['original_url']` since it's required

ursh/indico_ursh/controllers.py
```diff
@@ -52,8 +56,52 @@ def _process(self):
...
+    def _make_absolute_url(self, url):
+        return posixpath.join(config.BASE_URL, url[1:]) if url.startswith('/') else url
+
+    def _get_error_msg(self, response):
+        if response['status'] == 403:
+            return _('Shortcut already exists')
+
+    def _process_args(self):
+        super(RHCustomShortURLPage, self)._process_args()
+        from indico_ursh.plugin import UrshPlugin
```
ThiefMaster: keep imports at the beginning of the function

ursh/indico_ursh/controllers.py
```diff
@@ -52,8 +56,52 @@ def _process(self):
...
+        if not original_url or not shortcut:
+            raise BadRequest('Incomplete request for URL shortening service')
+        result = register_shortcut(original_url, shortcut)
+
+        error = result.get('error')
+        if not error:
+            kwargs = dict(success=True)
+        else:
+            kwargs = dict(success=False)
+            kwargs.update(msg=self._get_error_msg(result))
```
ThiefMaster: kwargs = {'success': False, 'msg': self._get_error_msg(result)}

ursh/indico_ursh/templates/ursh_dropdown.html
```diff
@@ -0,0 +1,23 @@
+<a href="" class="arrow js-dropdown ursh-dropdown {{ element_class }}" data-toggle="dropdown"
+   title="{% trans %}Short URL options{% endtrans %}"></a>
+<ul class="dropdown">
+    <li>
+        <a class="ursh-get"
+           data-original-url="{{ target }}">
+            {{ trans }}Obtain short URL{{ end_trans }}
+        </a>
+    </li>
+{%- if allow_create %}
```
ThiefMaster: that whole block should be indented

ursh/indico_ursh/templates/ursh_shortener_page.html
```diff
@@ -0,0 +1,35 @@
...
+    <p>
+        {% trans -%}
+            In this page, you can generate short links for any Indico URL.
+            Just enter the URL below and click "Shorten".
+        {%- endtrans %}
+    </p>
+    <div id="ursh-shorten-request-form" class="i-has-action">
+        <input id="ursh-shorten-input" name="ursh-shorten-original-url" type="url" style="font-family:monospace;"
+               spellcheck="false" autocomplete="off" autocapitalize="off"
+               placeholder="{% trans %}Enter an Indico URL to shorten{% endtrans %}...">
```
ThiefMaster: I'd keep the `...` inside the translated string (dunno if it's the same in all languages)

ThiefMaster:
The plugin is currently enabled on my dev instance if someone wants to play around with it instead of just looking at the code.

PR #73, opened 2019-06-01
Add Payfast (payfast.co.za) to Indico plugins

inus:
Support local South African payment gateway based on the PayPal paymen plugin for Indico


ThiefMaster:
Thanks a lot, but I don't think we want to maintain plugins for local payment services (as someone actively using it, you are a much better candidate for this!).

So IMHO it would be better if you kept the plugin in your own repo (NOT an indico-plugins fork but an `indico-plugin-payfast` repo that only contains your plugin).

pferreir:
Agreed. If the payment system were widely adopted, then it would be good candidate for integration into Indico's standard plugins. We have absolutely no way to maintain this ourselves, otherwise.

PR #74, opened 2019-06-14
Payment/Paypal: add warning about UTF-8 config

### LINKED ISSUE #60: PayPal: show warning about encoding
PayPal's API will send stuff encoded in windows 1252 by default, but mixed with UTF characters (go figure). The only way to reliable fix this issue is to have users configure their PayPal account to use UTF-8 by default. We should add a warning somewhere, linking to the docs, which should contain something like [this](https://stackoverflow.com/questions/12284341/paypal-ipn-override-charset/23593082#23593082).

pferreir:
Closes #60

### DIFF
--- a/payment_paypal/indico_payment_paypal/templates/event_settings_encoding_warning.html
+++ b/payment_paypal/indico_payment_paypal/templates/event_settings_encoding_warning.html
@@ -1,0 +1,9 @@
+{% from 'message_box.html' import message_box %}
+
+{% call message_box('warning', fixed_width=true) %}
+    {%- trans link='<a href="https://www.paypal.com/al/smarthelp/article/why-did-i-get-the-error-paypal-cannot-process-this-transaction-because-of-a-problem-with-the-seller%27s-website-ts1628" target="_blank">'|safe, endlink='</a>'|safe -%}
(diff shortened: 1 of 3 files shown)


payment_paypal/indico_payment_paypal/templates/event_settings_encoding_warning.html
```diff
@@ -0,0 +1,10 @@
+{% from 'message_box.html' import message_box %}
+
+{% call message_box('warning', fixed_width=true) %}
+    {%- trans -%}
+        Please make sure that you have your PayPal seller account configured to use <strong>UTF-8</strong> by default,
+        otherwise some transactions may end up not being acknowledged by Indico!<br>
+        You can find out how to set this up <a href="https://www.paypal.com/al/smarthelp/article/why-did-i-get-the-error-paypal-cannot-process-this-transaction-because-of-a-problem-with-the-seller%27s-website-ts1628">
```
ThiefMaster: I'd move the link to a trans param

pferreir: True.

PR #80, opened 2020-03-02
Fix height/width visual bugs in search tag

OmeGak:
## Before

Screen Shot 2020-03-02 at 13 41 22
Screen Shot 2020-03-02 at 13 41 33

## After

Screen Shot 2020-03-02 at 13 41 48
Screen Shot 2020-03-02 at 13 42 04

### DIFF
--- a/search/indico_search/client/main.scss
+++ b/search/indico_search/client/main.scss
@@ -169,11 +169,9 @@
 
 .search-tag {
     max-width: 112px;
-    display: table-cell;
+    display: flex;
     line-height: 13px;
-    height: 13px;
     vertical-align: middle;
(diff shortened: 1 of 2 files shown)


search/indico_search/client/main.scss
```diff
@@ -169,11 +169,9 @@
 
 .search-tag {
     max-width: 112px;
-    display: table-cell;
+    display: flexbox;
```
ThiefMaster: isn't that the legaxy flexbox (vs `display: flex` for the current one)?

OmeGak: True. Fixed.

PR #85, opened 2020-07-02
Vidyo: move room name sanitization to plugin

pferreir:
This is related to https://github.com/indico/indico/pull/4511/commits/286a2dc4e9b2feea1782ab19ece726d0ea54857b, which I'm including in https://github.com/indico/indico/pull/4511.

### DIFF
--- a/vc_vidyo/indico_vc_vidyo/forms.py
+++ b/vc_vidyo/indico_vc_vidyo/forms.py
@@ -8,12 +8,13 @@
 from __future__ import unicode_literals
 
 from wtforms.fields.core import BooleanField
-from wtforms.fields.simple import TextAreaField
+from wtforms.fields.simple import StringField, TextAreaField
 from wtforms.validators import DataRequired, Length, Optional, Regexp, ValidationError
 
 from indico.modules.vc.forms import VCRoomAttachFormBase, VCRoomFormBase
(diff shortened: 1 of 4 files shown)


vc_vidyo/indico_vc_vidyo/forms.py
```diff
@@ -7,19 +7,23 @@
...
 from indico.web.forms.base import generated_data
 from indico.web.forms.fields import IndicoPasswordField, PrincipalField
+from indico.web.forms.validators import IndicoRegexp
 from indico.web.forms.widgets import SwitchWidget
 
 from indico_vc_vidyo import _
 from indico_vc_vidyo.util import iter_user_identities, retrieve_principal
 
 
+ROOM_NAME_RE = re.compile(r'[\w\-]+')
```
ThiefMaster, line 26: Since `IndicoRegexp(ROOM_NAME_RE)` runs at import time as well, there's not much benefit in precompiling the regex instead of just passing the string directly to `IndicoRegexp()`

pferreir: I just moved it around. But yeah, I'll do that.

ThiefMaster:
Looks like Vidyo is no longer picky about room names so this isn't needed anymore :)

PR #88, opened 2020-10-30
Storage/S3: Use high-level API supporting chunks

### LINKED ISSUE #87: Allow chunked uploads to S3
Uploading files larger than 5GB fail otherwise (e.g. material packages).

We need to make sure that the MD5 checksum is correctly stored - AFAIK the ETag in case of chunked uploads is NOT the md5, so we should make sure to attach the `x-amz-meta-md5chksum` header (unless boto3 does this automatically).

ThiefMaster:
Otherwise it's impossible to store files larger than 5GB. We always include the header with the checksum now, because there is no harm in doing so and otherwise we'd have to first check how large the file we're storing is...

closes #87

### DIFF
--- a/storage_s3/indico_storage_s3/storage.py
+++ b/storage_s3/indico_storage_s3/storage.py
@@ -18,6 +18,7 @@ from io import BytesIO
 from tempfile import NamedTemporaryFile
 
 import boto3
+from boto3.s3.transfer import TransferConfig
 from botocore.config import Config
 from botocore.exceptions import ClientError
 from werkzeug.datastructures import Headers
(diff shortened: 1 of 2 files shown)


storage_s3/indico_storage_s3/storage.py
```diff
@@ -108,8 +109,18 @@ def _save(self, bucket, name, content_type, fileobj):
         checksum = get_file_checksum(fileobj)
         fileobj.seek(0)
         content_md5 = checksum.decode('hex').encode('base64').strip()
-        self.client.put_object(Body=fileobj, Bucket=bucket, Key=name,
-                               ContentType=content_type, ContentMD5=content_md5)
```
ThiefMaster, line 112: ContentMD5 is actually added automatically by botocore, so there's no need to manually pass it (and it wouldn't work for chunked uploads anyway since the md5 here is not the md5 on the individual chunks)

pferreir (approved):

PR #90, opened 2020-11-10
Zoom plugin

pferreir:
This adds the Zoom plugin developed by ENEA + CERN to the official Indico repo.

Some core changes needed for this plugin: https://github.com/indico/indico/pull/4697


vc_vidyo/indico_vc_vidyo/forms.py
```diff
@@ -8,12 +8,13 @@
 from __future__ import unicode_literals
```
ThiefMaster, line 8: this one as well

vc_vidyo/indico_vc_vidyo/plugin.py
```diff
@@ -7,6 +7,8 @@
```
ThiefMaster, line 7: and this one

vc_zoom/setup.py
```diff
@@ -0,0 +1,28 @@
...
+    description='Zoom video-conferencing plugin for Indico',
+    url='',
+    license='MIT',
+    author='Giovanni Mariano - ENEA',
+    author_email='giovanni.mariano@enea.it',
+    packages=find_packages(),
+    zip_safe=False,
+    include_package_data=True,
+    install_requires=[
+        'indico>=2',
```
ThiefMaster, line 18:
```suggestion
        'indico>=2.3.2.dev0',
```

vc_zoom/pytest.ini
```diff
@@ -0,0 +1,7 @@
+[pytest]
```
ThiefMaster, line 1: i changed some stuff in the pytest.ini files of other plugins in master (added our usual filterwarnings config) - maybe use that one already

vc_zoom/setup.py
```diff
@@ -0,0 +1,28 @@
...
+        'PyJWT'
```
ThiefMaster, line 19:
```suggestion
        'PyJWT<2'
```

vc_zoom/setup.py
```diff
@@ -0,0 +1,28 @@
+from __future__ import unicode_literals
+
+from setuptools import find_packages, setup
+
+
+setup(
+    name='indico-plugin-vc-zoom',
+    version='0.3-dev',
+    description='Zoom video-conferencing plugin for Indico',
+    url='',
```
ThiefMaster, line 10: link to the github repo (probably best to sync the setup.py content with another plugin)

vc_zoom/setup.py
```diff
@@ -0,0 +1,28 @@
+from __future__ import unicode_literals
+
+from setuptools import find_packages, setup
+
+
+setup(
+    name='indico-plugin-vc-zoom',
+    version='0.3-dev',
```
ThiefMaster, line 8:
```suggestion
    version='2.3-dev',
```

ThiefMaster (commented):
I'll have a look at the rest of the plugin tomorrow; just checked a few files so far

vc_zoom/indico_vc_zoom/controllers.py
```diff
@@ -29,3 +35,30 @@ def _process(self):
             flash(_("You are now the host of room '{room.name}'".format(room=self.vc_room)), 'success')
             result['success'] = True
         return jsonify(result)
+
+
+class RHWebhook(RH):
+    CSRF_ENABLED = False
+
+    def _check_access(self):
+        token = request.headers.get('Authorization')
```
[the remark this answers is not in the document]
pferreir: Nope, not bearer.

vc_zoom/indico_vc_zoom/util.py
```diff
@@ -40,8 +40,8 @@ def find_enterprise_email(user):
 
 
 def gen_random_password():
-    """Generate a random 10-character-long alphanumeric string."""
-    return ''.join(random.sample(string.ascii_lowercase + string.ascii_uppercase + string.digits, 10))
+    """Generate a random 10-character-long numeric string."""
+    return ''.join(random.sample(string.digits, 10))
```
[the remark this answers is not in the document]
pferreir: Good catch, thanks!

vc_zoom/pytest.ini
pferreir: Where exactly? In this repo?

vc_zoom/indico_vc_zoom/api/client.py
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    def __init__(self, api_key, api_secret, timeout=15):
+        """Create a new Zoom client.
+
+        :param api_key: the Zoom JWT API key
+        :param api_secret: the Zoom JWT API Secret
+        :param timeout: the time out to use for API requests
+        """
+        BASE_URI = "https://api.zoom.us/v2"
```
ThiefMaster, line 166: maybe move this to the class level

vc_zoom/indico_vc_zoom/api/client.py
```diff
@@ -0,0 +1,243 @@
...
+import time
+
+import jwt
+from requests import Session
+from requests.exceptions import HTTPError
+from pytz import utc
+
+
+def format_iso_dt(d):
+    """Convertdatetime objects to a UTC-based string.
```
ThiefMaster, line 19:
```suggestion
    """Convert a datetime object to a UTC-based string.
```

vc_zoom/indico_vc_zoom/api/client.py
```diff
@@ -0,0 +1,243 @@
...
+        return _handle_response(self.client.webinar.get(webinar_id))
+
+    def update_webinar(self, webinar_id, data):
+        return _handle_response(self.client.webinar.update(webinar_id, **data), 204, expects_json=False)
+
+    def delete_webinar(self, webinar_id):
+        return _handle_response(self.client.webinar.delete(webinar_id), 204, expects_json=False)
+
+    def check_user_meeting_time(self, user_id, start_dt, end_dt):
+        pass
```
ThiefMaster, line 234: TODO, missing or left empty on purpose?

vc_zoom/indico_vc_zoom/client/index.js
```diff
@@ -0,0 +1,34 @@
...
+
+    $.ajax({
+      url: $this.data('href'),
+      method: 'POST',
+      complete: IndicoUI.Dialogs.Util.progress(),
+    })
+      .done(function(result) {
+        if (handleAjaxError(result)) {
+          return;
+        } else {
```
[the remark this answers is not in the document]
ThiefMaster: ok nevermind, just saw the python code that returns exactly this with a normal 200 status code... what about using our `ExpectedError` exception for this which sends a proper error code (418)?

vc_zoom/indico_vc_zoom/forms.py
```diff
@@ -0,0 +1,100 @@
...
+                                           choices=[
+                                               ('everyone', _('Everyone')),
+                                               ('logged_in', _('Logged-in users')),
+                                               ('no_one', _("No one"))])
+
+
+class VCRoomForm(VCRoomFormBase):
+    """Contains all information concerning a Zoom booking."""
+
+    advanced_fields = {'mute_audio', 'mute_host_video', 'mute_participant_video'} | VCRoomFormBase.advanced_fields
```
ThiefMaster, line 38:
```suggestion
    advanced_fields = VCRoomFormBase.advanced_fields | {'mute_audio', 'mute_host_video', 'mute_participant_video'}
```

(dunno, feels almost a bit like a yoda condition to have it the ohter way round)

vc_zoom/indico_vc_zoom/forms.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @generated_data
+    def host(self):
+        if self.host_choice is None:
+            return None
+        return session.user.identifier if self.host_choice.data == 'myself' else self.host_user.data.identifier
+
+    def validate_host_user(self, field):
+        if not field.data:
+            raise ValidationError(_("Unable to find this user in Indico."))
```
ThiefMaster, line 100: When would that happen?

vc_zoom/indico_vc_zoom/notifications.py
```diff
@@ -0,0 +1,44 @@
...
+    template_module = get_template_module(
+        'vc_zoom:emails/notify_new_host.html',
+        plugin=ZoomPlugin.instance,
+        vc_room=vc_room,
+        actor=actor
+    )
+
+    new_host = principal_from_identifier(vc_room.data['host'])
+    email = make_email({new_host.email}, cc_list={actor.email}, template=template_module, html=True)
+    send_email(email, None, 'Zoom')
```
ThiefMaster, line 44: `send_email(email, module='Zoom')` but `module` is not used unless there's an event (no place to log it otherwise), so:

```suggestion
    send_email(email)
```

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -0,0 +1,442 @@
...
+            user == principal_from_identifier(room.data['host']) or
+            super(ZoomPlugin, self).can_manage_vc_room(user, room)
+        )
+
+    def _merge_users(self, target, source, **kwargs):
+        super(ZoomPlugin, self)._merge_users(target, source, **kwargs)
+        for room in VCRoom.query.filter(
+            VCRoom.type == self.service_name, VCRoom.data.contains({'host': source.identifier})
+        ):
+            room.data['host'] = target.id
```
ThiefMaster, line 410: Are you sure this is correct? You filter by `source.identifier` but set it to `target.id`

vc_zoom/indico_vc_zoom/util.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def find_enterprise_email(user):
+    """Find a user's first e-mail address which can be used by the Zoom API.
+
+    :param user: the `User` in question
+    :return: the e-mail address if it exists, otherwise `None`
+    """
+    from indico_vc_zoom.plugin import ZoomPlugin
+    providers = [auth.strip() for auth in ZoomPlugin.settings.get('email_domains').split(',')]
```
ThiefMaster, line 32: `providers` => `domains`?

vc_zoom/indico_vc_zoom/util.py
```diff
@@ -0,0 +1,101 @@
...
+    result = UserEmail.query.filter(
+        UserEmail.user == user,
+        ~User.is_blocked,
+        ~User.is_deleted,
+        db.or_(UserEmail.email.ilike("%%@{}".format(provider)) for provider in providers)
```
ThiefMaster, line 37:
```suggestion
        db.or_(UserEmail.email.ilike('%@{}'.format(provider)) for provider in providers)
```

vc_zoom/indico_vc_zoom/util.py
```diff
@@ -0,0 +1,101 @@
...
+    ).join(User).first()
+    return result.email if result else None
+
+
+def gen_random_password():
+    """Generate a random 10-character-long numeric string."""
+    return ''.join(random.choice(string.digits) for _ in range(8))
```
ThiefMaster, line 44: 8 != 10

vc_zoom/indico_vc_zoom/plugin.py
pferreir: Indeed. Bug. Well spotted.

vc_zoom/indico_vc_zoom/util.py
pferreir: this docstring is quite useless anyway 😆

vc_zoom/indico_vc_zoom/templates/emails/deleted.html
```diff
@@ -0,0 +1 @@
+{% extends 'vc/emails/deleted.html' %}
```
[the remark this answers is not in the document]
pferreir: Yeah, seems to work anyway.

vc_zoom/indico_vc_zoom/api/client.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+def _handle_response(resp, expected_code=200, expects_json=True):
+    resp.raise_for_status()
+    if resp.status_code != expected_code:
+        raise HTTPError("Unexpected status code {}".format(resp.status_code), response=resp)
+    if expects_json:
+        return resp.json()
+    else:
+        return resp
```
ThiefMaster, line 34:
```suggestion
    return resp.json() if expects_json else resp
```

However, if we never use the response in case of not expecting JSON, maybe returning `None` would be cleaner since that way it's perfectly clear that the response is not used at all?

vc_zoom/indico_vc_zoom/api/client.py
```diff
@@ -0,0 +1,243 @@
...
+
+        self.components = {
+            key: component(base_uri=BASE_URI, config=config, timeout=timeout)
+            for key, component in self._components.viewitems()
+        }
+
+    @property
+    def meeting(self):
+        """Get the meeting component."""
+        return self.components.get("meeting")
```
ThiefMaster, line 184:
```suggestion
        return self.components['meeting']
```

since we always have those components

vc_zoom/indico_vc_zoom/api/client.py
```diff
@@ -0,0 +1,243 @@
...
+
+    @property
+    def user(self):
+        """Get the user component."""
+        return self.components.get("user")
```
ThiefMaster, line 189:
```suggestion
        return self.components['user']
```

vc_zoom/indico_vc_zoom/api/client.py
```diff
@@ -0,0 +1,243 @@
...
+
+    @property
+    def webinar(self):
+        """Get the user component."""
+        return self.components.get("webinar")
```
ThiefMaster, line 194:
```suggestion
        """Get the webinar component."""
        return self.components['webinar']
```

vc_zoom/indico_vc_zoom/api/client.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+class ZoomIndicoClient(object):
+    def __init__(self):
+        self._refresh_client()
+
+    def _refresh_client(self):
```
ThiefMaster, line 201: not used anywhere else => move directly into `__init__`?

vc_zoom/indico_vc_zoom/client/index.js
```diff
@@ -0,0 +1,29 @@
...
+  document.querySelectorAll('.vc-toolbar .action-make-host').forEach(elem => {
+    elem.addEventListener('click', async () => {
+      let killProgress = IndicoUI.Dialogs.Util.progress();
+      try {
+        await indicoAxios.post(elem.dataset.href);
+        window.location.reload();
+      } catch (error) {
+        handleAxiosError(error);
+      }
+      killProgress();
```
ThiefMaster, line 26: I'd only do that inside catch since you reload after a successful POST anyway, so no need to "unblock" the page before that.

vc_zoom/indico_vc_zoom/client/index.js
```diff
@@ -0,0 +1,29 @@
...
+document.addEventListener('DOMContentLoaded', async () => {
+  $('.vc-toolbar').dropdown({
+    positioning: {
+      level1: {my: 'right top', at: 'right bottom', offset: '0px 0px'},
+    },
+  });
+
+  document.querySelectorAll('.vc-toolbar .action-make-host').forEach(elem => {
+    elem.addEventListener('click', async () => {
+      let killProgress = IndicoUI.Dialogs.Util.progress();
```
ThiefMaster, line 19:
```suggestion
      const killProgress = IndicoUI.Dialogs.Util.progress();
```

vc_zoom/indico_vc_zoom/client/index.js
```diff
@@ -0,0 +1,29 @@
...
+      try {
+        await indicoAxios.post(elem.dataset.href);
+        window.location.reload();
```
ThiefMaster, line 22: Does anything change on the page when making yourself host (besides maybe hiding this action)? If not we could just hide/remove it without reloading the page.

vc_zoom/indico_vc_zoom/forms.py
```diff
@@ -0,0 +1,110 @@
...
+            host = principal_from_identifier(defaults.host)
+            defaults.host_choice = 'myself' if host == session.user else 'someone_else'
+            defaults.host_user = None if host == session.user else host
+        super(VCRoomForm, self).__init__(*args, **kwargs)
+
+        if defaults.meeting_type is None:
+            del self.meeting_type
+        else:
+            for f in {self.mute_audio, self.mute_participant_video, self.waiting_room}:
+                f.validators = [HiddenUnless('meeting_type', 'regular')]
```
ThiefMaster, line 97: Using our `inject_validators` util here may be safer (note that this should be done before the super call)

vc_zoom/indico_vc_zoom/forms.py
```diff
@@ -0,0 +1,110 @@
...
+                                        ('regular', _('Regular Meeting')),
+                                        ('webinar', _('Webinar'))])
+
+    host_choice = IndicoRadioField(_("Meeting Host"), [DataRequired()],
+                                   choices=[('myself', _("Myself")), ('someone_else', _("Someone else"))])
+
+    host_user = PrincipalField(_("User"),
+                               [HiddenUnless('host_choice', 'someone_else'), DataRequired()])
+
+    password = StringField(_("Passcode"),
```
ThiefMaster, line 55: Should we use `passcode` instead of `password` for the internal names? Or is it `password` in the Zoom APIs as well?

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -0,0 +1,464 @@
...
+    join_before_host = BooleanField(_('Join Before Host'),
+                                    widget=SwitchWidget(),
+                                    description=_('Allow participants to join the meeting before the host starts the '
+                                                  'meeting. Only used for scheduled or recurring meetings.'))
+
+    waiting_room = BooleanField(_('Waiting room'),
+                                widget=SwitchWidget(),
+                                description=_('Participants may be kept in a waiting room by the host'))
+
+    zoom_phone_link = URLField(_('ZoomVoice phone number'),
```
ThiefMaster, line 90: "ZoomVoice" sounds a lot like a leftover from copying the field about "VidyoVoice" in the Vidyo plugin

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -0,0 +1,464 @@
...
+        self.connect(signals.event.times_changed, self._times_changed)
+        self.template_hook('event-vc-room-list-item-labels', self._render_vc_room_labels)
+        self.inject_bundle('main.js', WPSimpleEventDisplay)
+        self.inject_bundle('main.js', WPVCEventPage)
+        self.inject_bundle('main.js', WPVCManageEvent)
+
+    @property
+    def default_settings(self):
+        return dict(VCPluginMixin.default_settings, **{
+            'assistant_id': config.SUPPORT_EMAIL,
```
ThiefMaster, line 125: hmm not sure about using a default here... seems safer to really force the admin to enter something instead of defaulting to something that may or may not be the correct email, especially since they need to enter the tokens etc anyway

ThiefMaster: (when ditching this default the settings can go back to a class-level `default_settings = {...}` attribute again)

vc_zoom/indico_vc_zoom/util.py
```diff
@@ -0,0 +1,101 @@
...
+    :return: the e-mail address if it exists, otherwise `None`
+    """
+    from indico_vc_zoom.plugin import ZoomPlugin
+    providers = [auth.strip() for auth in ZoomPlugin.settings.get('email_domains').split(',')]
+    result = UserEmail.query.filter(
+        UserEmail.user == user,
+        ~User.is_blocked,
+        ~User.is_deleted,
+        db.or_(UserEmail.email.ilike("%%@{}".format(provider)) for provider in providers)
+    ).join(User).first()
```
ThiefMaster, line 38: Should we order this query so a primary email comes first? Just in case someone has multiple emails with Zoom accounts for whatever reasons...

vc_zoom/indico_vc_zoom/forms.py
pferreir: It's not a condition, though. Seems to me like a matter of personal preference.

vc_zoom/indico_vc_zoom/forms.py
pferreir: It's password in the API as well. That's why I left it.

vc_zoom/indico_vc_zoom/plugin.py
pferreir: Yeah, I forgot to check whether having something like this is still useful.

vc_zoom/indico_vc_zoom/cli.py
```diff
@@ -0,0 +1,39 @@
+# This file is part of the Indico plugins.
+# Copyright (C) 2020 CERN and ENEA
+#
+# The Indico plugins are free software; you can redistribute
+# them and/or modify them under the terms of the MIT License;
+# see the LICENSE file for more details.
+
+from __future__ import unicode_literals
```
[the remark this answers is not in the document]
pferreir: Hm... sure. But what is the advantage?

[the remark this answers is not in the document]
ThiefMaster: just to have less changes between 2.3-maintenance and master later and thus removing the risk of merge conflicts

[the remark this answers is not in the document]
pferreir: OK, I'll do that, then.

vc_zoom/indico_vc_zoom/client/index.js
pferreir: The details of the booking should also be updated (there's a host field in them).

vc_zoom/indico_vc_zoom/plugin.py
pferreir: Good point.

vc_zoom/indico_vc_zoom/api/client.py
```diff
@@ -0,0 +1,233 @@
...
+        :param timeout: the time out to use for API requests
+        """
+        # Setup the config details
+        config = {
+            "api_key": api_key,
+            "api_secret": api_secret
+        }
+
+        # Instantiate the components
+
```
ThiefMaster, line 172: 👾

vc_zoom/indico_vc_zoom/notifications.py
```diff
@@ -0,0 +1,44 @@
...
+
+    template_module = get_template_module(
+        'vc_zoom:emails/notify_start_url.html',
+        plugin=ZoomPlugin.instance,
+        vc_room=vc_room,
+        user=user
+    )
+
+    email = make_email(to_list, template=template_module, html=True)
+    send_email(email, None, 'Zoom')
```
ThiefMaster, line 29:
```suggestion
    send_email(email)
```

There's nothing to log without an event

vc_zoom/indico_vc_zoom/util.py
```diff
@@ -0,0 +1,102 @@
...
+def find_enterprise_email(user):
+    """Find a user's first e-mail address which can be used by the Zoom API.
+
+    :param user: the `User` in question
+    :return: the e-mail address if it exists, otherwise `None`
+    """
+    from indico_vc_zoom.plugin import ZoomPlugin
+    domains = [auth.strip() for auth in ZoomPlugin.settings.get('email_domains').split(',')]
+    # get all matching e-mails, primary first
+    result = UserEmail.query.filter(
```
ThiefMaster, line 34: I think you could use `db.session.query(UserEmail.email).***.scalar()` to get the email directly without the separate `result.email if result` step (but I haven't tried it). Nothing wrong with leaving it as it is though!

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -0,0 +1,455 @@
...
+
+        try:
+            settings = {
+                'host_video': not vc_room.data['mute_host_video'],
+            }
+
+            kwargs = {}
+            if is_webinar:
+                kwargs = {
+                    'type': 5 if scheduling_args else 6,
```
ThiefMaster, line 261: Looks like a good case for constants or an enum

ThiefMaster:
Something for before this gets merged: Maybe we could squash the first three commits from ENEA together into a single one, and also add the `VC/Zoom: ` prefix to all the commit messages that don't have it yet?

vc_zoom/indico_vc_zoom/util.py
pferreir: Doesn't `.scalar(...)` blow up if there are no matches?

pferreir: Never mind. Just read the docs.

pferreir: Yeah, works fine, thanks.

pferreir: Ah, actually, it blows up if one has > 1 e-mail which matches.

image

ThiefMaster: ah damn :/ `.limit(1)` might work though

pferreir: Yeah, but at this point I think the original version was easier to understand...

vc_zoom/indico_vc_zoom/api/client.py
```diff
@@ -0,0 +1,232 @@
...
+        return self.session.get("{}/users/{}/assistants".format(self.base_uri, user_id), params=kwargs)
+
+    def get(self, user_id, **kwargs):
+        return self.session.get("{}/users/{}".format(self.base_uri, user_id), params=kwargs)
+
+
+class ZoomClient(object):
+    """Zoom REST API Python Client."""
+
+    BASE_URI = "https://api.zoom.us/v2"
```
ThiefMaster, line 150: Should we make this overridable? AFAIK zoom provides some kind of deeper on-premise integration where you'd have a custom API host.

OTOH, if someone needs that they could send a PR since it's not relevant for us atm..

vc_zoom/indico_vc_zoom/util.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def find_enterprise_email(user):
+    """Find a user's first e-mail address which can be used by the Zoom API.
+
+    :param user: the `User` in question
+    :return: the e-mail address if it exists, otherwise `None`
+    """
+    from indico_vc_zoom.plugin import ZoomPlugin
+    domains = [auth.strip() for auth in ZoomPlugin.settings.get('email_domains').split(',')]
```
ThiefMaster, line 43:
```suggestion
    domains = [domain.strip() for domain in ZoomPlugin.settings.get('email_domains').split(',')]
```

vc_zoom/indico_vc_zoom/util.py
```diff
@@ -0,0 +1,113 @@
...
+    webinar = 5
+    recurring_webinar_no_time = 6
+    recurring_meeting_fixed_time = 9
+
+
+def find_enterprise_email(user):
+    """Find a user's first e-mail address which can be used by the Zoom API.
+
+    :param user: the `User` in question
+    :return: the e-mail address if it exists, otherwise `None`
```
ThiefMaster, line 40:
```suggestion
    :return: the e-mail address if it exists, otherwise ``None``
```

PR #92, opened 2020-12-01
VC/Zoom: update join URL when password changes

pferreir:


vc_zoom/indico_vc_zoom/forms.py
```diff
@@ -11,7 +11,7 @@
 
 from wtforms.fields.core import BooleanField, StringField
 from wtforms.fields.simple import TextAreaField
-from wtforms.validators import DataRequired, ValidationError
+from wtforms.validators import DataRequired
```
ThiefMaster, line 14: I don't think this change is correct - `_check_zoom_user` below uses this error!

pferreir: Whoops. Rebase problem.

vc_zoom/tests/operation_test.py
```diff
@@ -128,3 +126,42 @@ def test_host_change(create_user, mocker, create_meeting, zoom_plugin, zoom_api,
...
+                'host_video': True,
+                'mute_upon_entry': False,
+                'participant_video': True,
+                'waiting_room': False
+            }
+        }
+        _get_meeting.called = True
+        return result
+    _get_meeting.called = False
+    mocker.patch('indico_vc_zoom.plugin.ZoomIndicoClient.get_meeting', _get_meeting)
```
ThiefMaster, line 158: AFAIK this returns a Mock object, so you can use that to assert it has been called instead of doing the `_get_meeting.called` stuff manually above.

vc_zoom/tests/operation_test.py
```diff
@@ -128,3 +126,42 @@ def test_host_change(create_user, mocker, create_meeting, zoom_plugin, zoom_api,
...
+
+    zoom_plugin.update_room(vc_room, vc_room.events[0].event)
+
+    assert zoom_api['update_meeting'].call_args == (('12345abc', {
```
ThiefMaster, line 162: There's something like `.assert_called_with()` on Mock objects

vc_zoom/tests/operation_test.py
pferreir: That's not the goal of `.called`. I'm changing the internal behaviour of the mock, not asserting whether it was called or not.

ThiefMaster: oh indeed, didn't notice that

vc_zoom/tests/operation_test.py
pferreir: Good idea, thanks.

PR #93, opened 2020-12-01
Support getting users via authenticators

ThiefMaster:


vc_zoom/indico_vc_zoom/util.py
```diff
@@ -33,22 +34,63 @@ class ZoomMeetingType(RichIntEnum):
...
+        domain = ZoomPlugin.settings.get('enterprise_domain')
+        for username in _iter_user_identifiers(user):
+            yield '{}@{}'.format(username, domain)
+
+
+@memoize_request
+def find_enterprise_email(user):
+    """Get the email address of a user that has a zoom account."""
+    client = ZoomIndicoClient()
+    return next((email for email in _iter_user_emails(user) if client.get_user(email, silent=True)), None)
```
pferreir, line 93: So, we will always do at least a GET request to confirm the user exists? It shouldn't be a big deal if we're only using this at create/update time, but still just wondering if we cannot assume the e-mail we found is OK and then fail later?

vc_zoom/tests/operation_test.py
```diff
@@ -107,6 +107,7 @@ def test_room_creation(create_meeting, zoom_api):
 
 
 def test_host_change(create_user, mocker, create_meeting, zoom_plugin, zoom_api, request_context):
+    mocker.patch('indico_vc_zoom.plugin.find_enterprise_email', return_value='joe.bidon@megacorp.xyz')
```
pferreir, line 110: This kind of keeps `find_enterprise_email` from being covered. Maybe we should add a separate test for it in that case?

pferreir (approved):

vc_zoom/indico_vc_zoom/util.py
ThiefMaster: at least if there are multiple emails (in case of domains) that won't work.. but if there's exactly one then this would indeed work. the disadvantage is that we need to exhaust the user emails iterator which can also send some unnecessary requests (in case multiple identity providers are used, since we search for the user's email on each of them)

ThiefMaster: i think during creation/update we call `find_enterprise_email` at least once anyway and since it's memoized now it will just be a single zoom api request. and if there is only one possible email for the user, then there won't be any extra requests ;)

pferreir: Sure, makes sense. It will never be an enormous performance penalty. I was just wondering if just taking the first result and expecting that would work wouldn't be good enough:tm:.

pferreir: But yeah, in any case we don't lose much doing it the right way.

vc_zoom/tests/operation_test.py
ThiefMaster: Good point, I'll add one later. Not super straightforward to write a test for it because of all the dependencies that need to be mocked out.

PR #94, opened 2020-12-02
VC/Zoom: set room recurrence when cloning event

pferreir:
related: indico/indico#4732

### DIFF
--- a/vc_zoom/indico_vc_zoom/plugin.py
+++ b/vc_zoom/indico_vc_zoom/plugin.py
@@ -444,6 +444,25 @@ class ZoomPlugin(VCPluginMixin, IndicoPlugin):
                 self.logger.error("Can't delete room")
                 raise VCRoomError(_('Problem deleting room'))
 
+    def clone_room(self, old_event_vc_room, link_object):
+        vc_room = old_event_vc_room.vc_room
+        is_webinar = vc_room.data.get('meeting_type', 'regular') == 'webinar'
+        has_only_one_association = len({assoc.event_id for assoc in vc_room.events}) == 1
+        new_assoc = super(ZoomPlugin, self).clone_room(old_event_vc_room, link_object)
+
+        if has_only_one_association:
+            update_zoom_meeting(vc_room.data['zoom_id'], {
+                'start_time': None,
+                'duration': None,
+                'type': (
+                    ZoomMeetingType.recurring_webinar_no_time
+                    if is_webinar
+                    else ZoomMeetingType.recurring_meeting_no_time
+                )
+            })
+
+        return new_assoc
+
     def get_blueprints(self):
         return blueprint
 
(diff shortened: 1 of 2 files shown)


vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -438,6 +438,25 @@ def delete_room(self, vc_room, event):
                 self.logger.error("Can't delete room")
                 raise VCRoomError(_('Problem deleting room'))
 
+    def clone_room(self, old_event_vc_room, link_object):
+        vc_room = old_event_vc_room.vc_room
+        is_webinar = vc_room.data.setdefault('meeting_type', 'regular') == 'webinar'
```
ThiefMaster, line 443: setdefault or get?

pferreir: Indeed. Copy-paste.

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -438,6 +438,25 @@ def delete_room(self, vc_room, event):
                 self.logger.error("Can't delete room")
                 raise VCRoomError(_('Problem deleting room'))
 
+    def clone_room(self, old_event_vc_room, link_object):
+        vc_room = old_event_vc_room.vc_room
+        is_webinar = vc_room.data.get('meeting_type', 'regular') == 'webinar'
+        has_only_one_association = len(vc_room.events) == 1
```
ThiefMaster, line 444: I guess for existing events we can just fix it manually (since this wouldn't trigger during future clones since there are already more links).. I checked and there are a few such cases in prod.

```python
from indico_vc_zoom.api import ZoomIndicoClient
from indico_vc_zoom.util import ZoomMeetingType

client = ZoomIndicoClient()
for vc_room in VCRoom.query.filter(VCRoom.type == 'zoom').options(joinedload('events')).all():
    if len(vc_room.events) <= 1:
        continue
    meeting_type = client.get_meeting(vc_room.data['zoom_id'])['type']
    is_webinar = vc_room.data.get('meeting_type', 'regular') == 'webinar'
    if meeting_type in (ZoomMeetingType.scheduled_meeting, ZoomMeetingType.webinar):
        print('Fixing type of {} ({})'.format(vc_room, vc_room.data['zoom_id']))
        update_zoom_meeting(vc_room.data['zoom_id'], {
            'start_time': None,
            'duration': None,
            'type': (
                ZoomMeetingType.recurring_webinar_no_time
                if is_webinar
                else ZoomMeetingType.recurring_meeting_no_time
            )
```

ThiefMaster: Actually, `events` is a bit misleading; it's actually all links. So right now it would also not work properly in case the room is linked to multiple things within the same event. We should probably check the size of `{assoc.event_id for assoc in vc_room.events}`

ThiefMaster: I think the `association_is_new = room_assoc.vc_room is None` logic when associating an existing room manually has the same bug.

pferreir: Very good catch!

ThiefMaster:
changelog entry in README.md, 2.3b2 section, would be great ;)

pferreir:
> changelog entry in README.md, 2.3b2 section, would be great ;)

Done!

ThiefMaster (approved):

PR #100, opened 2020-12-10
Add CI workflow for github actions

### LINKED ISSUE #91: Move CI to GitHub actions instead of Travis
We can probably copy a lot of the general setup steps from Indico.

Some ideas for better performance:

- install the venv and indico in a setup step (so it gets cached)
- plugin node modules are relatively small, no need to cache those (would likely be messy since they are plugin-specific)
- like on travis, have separate parallel jobs for the plugins that need testing (no need to split lint/testing like we do in indico; both these things are much faster for plugins anyway)

plourenco:
Closes #91 

- [ ] Use different steps for setup?


ThiefMaster:
Maybe we could even do that in 2.3-maintenance so we can already ditch Travis there as well? It's not very different anyway besides the python version.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,62 @@
...
+        shell: bash
+        run: echo "BRANCH_NAME=$(echo ${GITHUB_HEAD_REF} | sed 's/.*\///')" >> $GITHUB_ENV
+      - name: Install python dependencies
+        if: steps.cache-pip.outputs.cache-hit != 'true'
+        run: |
+          python3 -m venv .venv
+          source .venv/bin/activate
+          pip install -U pip setuptools
+          pip install wheel
+          pip install -e git+https://github.com/indico/indico.git@${BRANCH_NAME}#egg=indico
```
ThiefMaster, line 47: I think for testing we need to hardcode the branch here, since the push runs in your repo currently fail due to upstream not having a branch named `ci`.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Activate virtualenv for later steps
+        run: |
+          echo "VIRTUAL_ENV=$(pwd)/.venv" >> $GITHUB_ENV
+          echo "$(pwd)/.venv/bin" >> $GITHUB_PATH
+      - name: Install
+        env:
+          PLUGIN: ${{ matrix.plugin }}
```
ThiefMaster, line 54: Can't you use `${{ matrix.plugin }}` directly in the `run` command below?

plourenco: I tested it on a separate branch and couldn't do it myself. I'm not sure if this is possible.
https://github.com/plourenco/indico-plugins/runs/1530852142?check_suite_focus=true (expand line 1)

plourenco: I couldn't do it previously, but the mistake was mine. I'm going to try that again.

.github/workflows/ci.yml
plourenco: Yes, I agree. I made a few tests previously on a separate branch. But as soon as I finish we can run with `ci` once more.

PR #101, opened 2020-12-10
Zoom: remove assistants, disallow host changes, add "make me alternative host"

pferreir:
Assistants are basically useless due to the rate limits in the Zoom API.

To be run right after install:

```python
for room in VCRoom.query.filter(VCRoom.type == 'zoom'):
    try:
        ZoomPlugin.instance.refresh_room(room, None)
    except Exception, e:
        print room, e
```


pferreir:
(Let's wait for confirmation before merging this)

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -501,6 +462,12 @@ def _merge_users(self, target, source, **kwargs):
         ):
             room.data['host'] = target.identifier
             flag_modified(room, 'data')
+        for room in VCRoom.query.filter(
+            VCRoom.type == self.service_name, VCRoom.data.contains({'alternative_hosts': [source.identifier]})
```
ThiefMaster, line 466: Will this contains filterwork when there are already multiple `alternative_hosts`, ie is it a deep contains or does it only check the top-level json object for containment, and then for an exact value when it comes to the list in there?

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -35,15 +34,16 @@
 from indico_vc_zoom.blueprint import blueprint
 from indico_vc_zoom.cli import cli
 from indico_vc_zoom.forms import VCRoomAttachForm, VCRoomForm
-from indico_vc_zoom.notifications import notify_new_host, notify_host_start_url
-from indico_vc_zoom.util import (UserLookupMode, fetch_zoom_meeting, find_enterprise_email, gen_random_password,
-                                 get_schedule_args, get_url_data_args, update_zoom_meeting, ZoomMeetingType)
+from indico_vc_zoom.notifications import notify_host_start_url
+from indico_vc_zoom.util import (UserLookupMode, fetch_zoom_meeting, find_enterprise_email, get_alt_host_emails,
+                                 gen_random_password, get_schedule_args, get_url_data_args, process_alternative_hosts,
+                                 update_zoom_meeting, ZoomMeetingType)
```
ThiefMaster, line 40: isort may want to change the order here ;)

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -395,6 +350,11 @@ def update_room(self, vc_room, event):
         if vc_room.data['mute_host_video'] == zoom_meeting_settings['host_video']:
             changes.setdefault('settings', {})['host_video'] = not vc_room.data['mute_host_video']
 
+        alternative_hosts = process_alternative_hosts(zoom_meeting_settings.get('alternative_hosts', []))
```
plourenco, line 353: Isn't the default value of `zoom_meeting_settings.get('alternative_hosts')` supposed to be a string to be parsed in `process_alternative_hosts`?

ThiefMaster: I think `zoom_meeting_settings` is what we store in the indico db (VCRoom.data), and the comma-separated string is what's used on the Zoom API side.

vc_zoom/indico_vc_zoom/plugin.py
ThiefMaster: ok, not only here.. `isort -rc vc_zoom/indico_vc_zoom/` fixes a few other files as well

vc_zoom/indico_vc_zoom/templates/info_box.html
```diff
@@ -11,6 +12,17 @@
         <dt>{% trans %}Host{% endtrans %}</dt>
         <dd>{{ (host|decodeprincipal).full_name }}</dd>
     {% endif %}
+    {% if alt_hosts %}
+        <dt>{% trans %}Alternative host(s){% endtrans %}</dt>
+        <dd>
+            {% for alt_host in alt_hosts %}
+                <span>{{ alt_host|decodeprincipal}}</span>
+                {% if not loop.last %}
+                ,
```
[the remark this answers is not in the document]
pferreir: It does return an object. I had fixed that but forgot to push. Thanks!

vc_zoom/indico_vc_zoom/plugin.py
pferreir: I did test the query and it finds rooms also with more than one alternative host.

vc_zoom/indico_vc_zoom/plugin.py
pferreir: Indeed. I confused it with the data which gets stored in the room's data. Thanks!

vc_zoom/indico_vc_zoom/templates/info_box.html
```diff
@@ -11,6 +12,17 @@
...
         <dd>{{ (host|decodeprincipal).full_name }}</dd>
     {% endif %}
+    {% if alt_hosts %}
+        <dt>
+            {% trans count=alt_hosts|length -%}
+                Alternative host
+            {%- pluralize -%}
+                Alternative hosts
+            {%- endtrans %}
+        <dd>
```
ThiefMaster, line 22:
```suggestion
        </dt>
        <dd>
```

vc_zoom/indico_vc_zoom/templates/manage_event_info_box.html
```diff
@@ -12,6 +13,17 @@
...
         {{ (host|decodeprincipal).full_name }}
     </dd>
+    {% if alt_hosts %}
+        <dt>
+            {% trans count=alt_hosts|length -%}
+                Alternative host
+            {%- pluralize -%}
+                Alternative hosts
+            {%- endtrans %}
+        <dd>
```
ThiefMaster, line 23:
```suggestion
        </dt>
        <dd>
```

PR #71, opened 2019-03-18
Vidyo: Post-process event metadata

pferreir:
Related to https://github.com/indico/indico/pull/3840

PR #103, opened 2021-01-11
Include Zoom join URL(s) in ICAL files

ThiefMaster:
We can most likely use the same APIs as in #71 for this.

Important: This must not expose private join links (with the passcode included) to people who would otherwise not have access. We may need to include the user in the `metadata_postprocess` signal since in case of a call through the HTTP API we do not have a `session.user`.


vc_zoom/README.md
```diff
@@ -27,6 +27,7 @@
 - Do not allow passcodes that are too long for zoom
 - Remove the "Assistant Zoom ID" logic due to problems with Zoom's API rate limits (all meetings created were counted against the assistant's rate limit instead of the host's); this means the host can no longer be changed, but Indico instead provides an option to event managers to make themselves a co-host.
 - Fix an error when changing the linked object of a recurring Zoom room in Indico
+- Include Zoom join links in the event's ical export (note: only Zoom meetings with a public passcode are displayed unless you have at least Indico v2.3.3)
```
ThiefMaster, line 30: The 2.3.3 requirement is due to indico/indico@4a456df334

plourenco (approved):

PR #104, opened 2021-01-19
Create a dummy search plugin

plourenco:
Moves the dummy data from indico/indico#4736 into a new plugin as a proof of concept for an Indico search provider.


dummy_search/setup.py
```diff
@@ -0,0 +1,32 @@
...
+# Copyright (C) 2014 - 2021 CERN
+#
+# The CERN Indico plugins are free software; you can redistribute
+# them and/or modify them under the terms of the MIT License; see
+# the LICENSE file for more details.
+
+from setuptools import find_packages, setup
+
+
+setup(
```
ThiefMaster, line 11: Let's also use the new structure - see `vc_zoom/setup.{py,cfg}` for an example. Also, version should be 3.0-dev.

dummy_search/src/plugin.py
```diff
@@ -0,0 +1,42 @@
...
+from indico.modules.search.base import IndicoSearchProvider, SearchTarget
+
+
+class DummySearchPlugin(IndicoPlugin):
+    def init(self):
+        super(DummySearchPlugin, self).init()
+        self.connect(signals.get_search_providers, self.get_search_providers)
+
+    def get_search_providers(self, sender, **kwargs):
+        yield TestSearch
```
ThiefMaster, line 24: you can `return` if it's just a single one

dummy_search/setup.py
plourenco: Good idea, but shouldn't this be based on 2.3 instead, as the search interface?

ThiefMaster: That's what I was saying during our meeting - with 3.0 progressing well and deployment being no longer an issue (I originally thought it might require much deeper infrastructure changes, like moving to CentOS8), we can ditch the plan of keeping the new search compatible with 2.3 and just rebase it on top of master.

And for the core all that needs to be done to use the new search logic on Python 3 is removing the now obsolete `__future__` imports from the files in ` indico/modules/search`

plourenco: My bad, I recall that - completely forgot this PR meant to include that too.

ThiefMaster:
I don't think we really need this anymore... at least not as something to be eventually merged into the repo

plourenco:
Re-opening as this will become useful as a reference implementation for an external search plugin backed up by an arbitrary service, as we will suggest in the new documentation rising in #4922.

PR #110, opened 2021-04-09
Livesync: Include subcontributions in the initial export

plourenco:
🚗💨

### DIFF
--- a/livesync/indico_livesync/initial.py
+++ b/livesync/indico_livesync/initial.py
@@ -119,7 +119,7 @@ def query_subcontributions():
         .join(Contribution.event.of_type(contrib_event))
         .outerjoin(Contribution.session.of_type(contrib_session))
         .outerjoin(Contribution.session_block.of_type(contrib_block))
-        .filter(~SubContribution.is_deleted & ~Contribution.is_deleted & ~contrib_event.is_deleted)
+        .filter(~SubContribution.is_deleted, ~Contribution.is_deleted, ~contrib_event.is_deleted)
         .options(
             selectinload(SubContribution.person_links),
             contrib_strategy,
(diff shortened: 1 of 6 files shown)


livesync/indico_livesync/initial.py
```diff
@@ -89,6 +89,48 @@ def query_contributions():
...
+    session_block_strategy.joinedload(contrib_block.own_venue)
+    session_block_strategy.joinedload(contrib_block.own_room).options(raiseload('*'), joinedload('location'))
+
+    return (
+        SubContribution.query
+        .join(Contribution)
+        .join(Contribution.event.of_type(contrib_event))
+        .outerjoin(Contribution.session.of_type(contrib_session))
+        .outerjoin(Contribution.session_block.of_type(contrib_block))
+        .filter(~SubContribution.is_deleted & ~Contribution.is_deleted & ~contrib_event.is_deleted)
```
ThiefMaster, line 122:
```suggestion
        .filter(~SubContribution.is_deleted, ~Contribution.is_deleted, ~contrib_event.is_deleted)
```

no technical difference but I think we usually do is that way for simple AND-linked criteria

livesync/indico_livesync/uploader.py
```diff
@@ -48,17 +48,17 @@ def run(self, records):
             self.processed_records(batch)
         self.logger.info('%s finished', self_name)
 
-    def run_initial(self, events, total=None):
+    def run_initial(self, records, total=None):
```
ThiefMaster, line 51: I think we can make `total` mandatory..

plourenco: Previously I would say we could want to run an export without progress. 
But right now, I don't see any cases for making it optional either.

ThiefMaster: yeah, I think if we ever find a reason why we wouldn't want a progress display, that should be made explicit through a new CLI arg. but it doesn't seem particularly useful...

PR #113, opened 2021-04-27
LiveSync & Citadel: Handle attachments and file uploads during queue runs

ThiefMaster:


livesync/indico_livesync/uploader.py
```diff
@@ -56,7 +54,7 @@ def run_initial(self, records, total):
             ((rec, SimpleChange.created) for rec in records),
             total,
             lambda entry: entry[0].id,
```
plourenco, line 56: It was me changing this earlier but I just saw we can switch to the same method
```suggestion
            lambda entry: getattr(entry[0], 'id'),
```

ThiefMaster: uh why? getattr with a static attr and no default isn't really useful

citadel/indico_citadel/cli.py
```diff
@@ -19,12 +25,66 @@ def cli():
...
     agent = LiveSyncAgent.query.filter(LiveSyncAgent.backend_name == 'citadel').first()
     if agent is None:
         print('No citadel livesync agent found')
         return
+    if not CitadelSearchAppIdMap.query.has_rows():
+        print('It looks like you did not export any data to Citadel yet.')
+        print(cformat('To do so, run %{yellow!}indico livesync initial-export {}%{reset}').format(agent.id))
+        return
+
     backend = agent.create_backend()
```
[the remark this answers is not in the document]
ThiefMaster: the error would still need to be handled in the cli

livesync/indico_livesync/uploader.py
plourenco: Never mind, I had the impression your intention was to make it more efficient while repeating this multiple times

PR #118, opened 2021-04-30
Citadel: Use new result schemas

ThiefMaster:
related to indico/indico#4888


citadel/indico_citadel/result_schemas.py
```diff
@@ -0,0 +1,95 @@
...
+from indico.modules.search.base import SearchTarget
+from indico.modules.search.result_schemas import (AggregationSchema, AttachmentResultSchema, BucketSchema,
+                                                  EventResultSchema, ResultItemSchema, ResultSchema)
+
+
+class CitadelEventResultSchema(EventResultSchema):
+    @pre_load
+    def _translate_keys(self, data, **kwargs):
+        data = data.copy()
+        data['event_type'] = data.pop('type_format')
```
plourenco, line 22: What I would suggest is instead of having this schema, changing the name on the Indico side to `event_type`.
Within citadel, we can change the `EventRecordSchema` to:
```
type_format = fields.String(attribute='_type.name')
```

ThiefMaster: yeah, i plan to do that ;) I'm also thinking of moving the dump schemas to the livesync plugin since right now they reside in indico but they are only used by livesync - and IMO it's more logical to have the result schemas in indico (since they describe the format search plugins need to use to send data back to indico) but the dump schemas in livesync..

plourenco: In the beginning, I would be strongly against that as I intended to use Indico as a single source of truth - providing the schemas and then re-using them as a validation method. For example: 
```
class EventSchema(mm.Schema):
    event_id = fields.Int(attribute='id')
```

And then dumping a result with `EventSchema().dump({'id': 1})` and then adapting it to validate or load in the same way with `EventSchema().validate({'event_id': 1}). But now that we intend to have two schemas for in and out, I'm not so sure anymore.

PR #119, opened 2021-05-03
Citadel: Integrate filters with Indico and update placeholders

plourenco:
Even though we don't need to pass the filters into Indico, as most of them come directly within the result aggregations, having them available is useful for providing i18n or getting their name from the query string.

Having a pre-defined list of them available through Indico or the search plugin is discussible. Even though I think it makes more sense the latter, as this PR, since they should change between implementations.

### DIFF
--- a/citadel/indico_citadel/search.py
+++ b/citadel/indico_citadel/search.py
@@ -64,10 +64,10 @@ placeholders = {
     'person': ('_data.persons.name', _("A speaker, author or event chair's name")),
     'affiliation': ('_data.persons.affiliation', _("A speaker, author or event chair's affiliation")),
     'type': ('_data.type', _('An entry type (such as: conference, meeting, link, file)')),
-    'venue': ('_data.location.venue_name', _("The entry's venue name")),
-    'room': ('_data.location.room_name', _("The entry's room name")),
-    'address': ('_data.location.address', _("The entry's address")),
-    'file': ('_data.filename', _("An attachment's file name")),
+    'venue': ('_data.location.venue_name', _("Name of the venue")),
+    'room': ('_data.location.room_name', _("Name of the room")),
+    'address': ('_data.location.address', _("Address of the venue")),
+    'file': ('_data.filename', _("Name of the attached file")),
     'keyword': ('_data.keywords', _('A keyword associated with an event'))
 }
 
@@ -80,6 +80,6 @@ filters = {
     'person': _('Person'),
     'type_format': _('Type'),
     'venue': _('Location'),
-    'start_range': _('Start Date'),
+    'start_range': _('Date'),
     'category': _('Category')
 }
--- a/citadel/indico_citadel/search.py
+++ b/citadel/indico_citadel/search.py
@@ -9,7 +9,7 @@ import requests
 from requests.exceptions import RequestException
 from werkzeug.urls import url_join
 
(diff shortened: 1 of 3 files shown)


citadel/indico_citadel/search.py
```diff
@@ -53,27 +53,33 @@ def search(self, query, access, page=1, object_types=(), **params):
...
+        return [SearchFilter(key, label) for key, label in filters.items()]
 
 
 placeholders = {
     'title': ('_data.title', _('An entry title (such as: event, contribution, attachment, event_note)')),
     'person': ('_data.persons.name', _("A speaker, author or event chair's name")),
     'affiliation': ('_data.persons.affiliation', _("A speaker, author or event chair's affiliation")),
     'type': ('_data.type', _('An entry type (such as: conference, meeting, link, file)')),
-    'venue': ('_data.location.venue_name', _("The event's venue name")),  # TODO: multiple locations
+    'venue': ('_data.location.venue_name', _("The entry's venue name")),
```
ThiefMaster, line 67: What about rewording these titles like this?

> Name of the venue
Name of the room
Address of the venue
Name of the attached file

Especially the "entry" sounds a bit strange for end users

citadel/indico_citadel/search.py
```diff
@@ -53,27 +53,33 @@ def search(self, query, access, page=1, object_types=(), **params):
...
-    'affiliation': 'Affiliation',
-    'person': 'Person',
-    'type_format': 'Type',
-    'venue': 'Location',
-    'start_range': 'Start Date',
+    'affiliation': _('Affiliation'),
+    'person': _('Person'),
+    'type_format': _('Type'),
+    'venue': _('Location'),
+    'start_range': _('Start Date'),
```
ThiefMaster, line 83: Isn't this used for files/notes as well now? In that case I'd just use "Date"

citadel/indico_citadel/util_test.py
```diff
@@ -18,7 +18,7 @@
     ('hey title:something', '+title:something +(hey)'),
     ('title:something hey', '+title:something +(hey)'),
     ('hey title:something hey person:john', '+title:something +person:john +(hey hey)'),
-    ('<*\\^()', '+(\\<\\*\\\\\\^\\(\\))'),
+    ('<*\\^()', '+(\\<*\\\\\\^\\(\\))'),
```
ThiefMaster, line 21: Maybe we could add another more "realistic" case here? Like searching for `foo*` to make it clear it not being escaped is intended and not a mistake?

citadel/indico_citadel/search.py
plourenco: Sounds good. I've renamed all of the listed above.

citadel/indico_citadel/search.py
```diff
@@ -9,7 +9,7 @@
 from requests.exceptions import RequestException
 from werkzeug.urls import url_join
 
-from indico.modules.search.base import IndicoSearchProvider, SearchPlaceholder
+from indico.modules.search.base import IndicoSearchProvider, SearchFilter
```
ThiefMaster, line 12: I think the commit for this change in indico itself isn't pushed anywhere

plourenco: I was pending on this PR, but it's there now 👍

PR #121, opened 2021-05-04
Citadel: Strip HTML from descriptions/notes

ThiefMaster:


plourenco (approved):
Can this become a problem with markdown in notes later on?

ThiefMaster:
I don't think so - stripping HTML tags from markdown shouldn't do anything problematic to it.

PR #122, opened 2021-05-05
Use all worker threads all the time

ThiefMaster:
The previous thread pool implementation launched n threads but then waited for all of them to finish before launching the next n. This one processes the data constantly instead of waiting for a full batch to finish.

### DIFF
--- a/citadel/indico_citadel/cli.py
+++ b/citadel/indico_citadel/cli.py
@@ -42,11 +42,13 @@ def upload(batch, force, max_size):
         print('Citadel is not properly configured.')
         return
 
-    total, errors = backend.run_export_files(batch, force, max_size=max_size)
-    if not errors:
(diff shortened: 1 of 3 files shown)


citadel/indico_citadel/cli.py
```diff
@@ -42,11 +42,13 @@ def upload(batch, force, max_size):
...
-    if not errors:
+    total, errors, aborted = backend.run_export_files(batch, force, max_size=max_size)
+    if not errors and not aborted:
         print(f'{total} files uploaded')
         backend.set_initial_file_upload_state(True)
         db.session.commit()
     else:
+        if aborted:
+            print('Upload aborted')
         print(f'{total} files uploaded, but {errors} failed')
```
ThiefMaster, line 53: TODO the `total` value here is nonsense in case of failures/aborts since it comes from the query, not the actually-processed records

PR #124, opened 2021-05-06
Refactor the search query parser

plourenco:
Escape everything except for valid placeholders. Also ensures the original keyword positioning isn't changed.

The old logic splitter placeholders and keywords:
Input: `hey title:john hey`
Ouput: `+title:john +(hey hey)`

Input: `title:john AND "developers meeting"`
Output: `+title:john +(AND "developers meeting")` -- Error

The new logic ensures the content and positioning is preserved only doing light parsing for escaping and whitelisting placeholders:
Input: `hey title:john hey invalid_placeholder:foo`
Output: `hey +title:john hey invalid_placeholder\\:foo`

Input: `title:john AND "developers meeting"`
Output: `+title:john AND "developers meeting"`

Making sure each of the keywords is required in the output will be left up to Elastic search using `default_operator=AND` if necessary.

### DIFF
--- a/citadel/indico_citadel/util.py
+++ b/citadel/indico_citadel/util.py
@@ -67,15 +67,20 @@ def format_query(query, placeholders):
     :param placeholders: placeholder whitelist
     :returns escaped query
     """
-    patt = r'({}):([^:"\s]+|"[^"]+")\s*'.format('|'.join(placeholders.keys()))
-    # Extract all placeholders
-    p = [f'+{placeholders[x.group(1)]}:{x.group(2)}'
-         for x in re.finditer(patt, query) if x.group(1) in placeholders]
-    # Escape keyword based arguments
-    query = escape(re.sub(patt, '', query)).strip()
-    if query:
-        p.append(f'+({query})')
-    return ' '.join(p)
+    patt = r'(?:^|\s)({}):([^:"\s]+|"[^"]+")(?:$|\s)'.format('|'.join(placeholders.keys()))
+    idx = 0
+    keys = []
+    for x in re.finditer(patt, query):
+        placeholder = f'+{placeholders[x.group(1)]}:{escape(x.group(2))}'
+        if idx != x.start():
+            keys.append(escape(query[idx:x.start()]))
+        keys.append(placeholder)
+        idx = x.end()
+
+    if idx != len(query):
+        keys.append(escape(query[idx:len(query)]))
+
+    return ' '.join(keys)
 
 
 def format_filters(params, filters, range_filters):
--- a/citadel/indico_citadel/util_test.py
+++ b/citadel/indico_citadel/util_test.py
@@ -12,16 +12,18 @@ from indico_citadel.util import format_query, remove_none_entries
 
 @pytest.mark.parametrize(('query', 'expected'), [
     ('title:"my event" ola person:john ola some:yes ola',
-     '+title:"my event" +person:john +(ola ola some\\:yes ola)'),
-    ('title:"my title:something"', '+title:"my title:something"'),
-    ('hello', '+(hello)'),
-    ('hey title:something', '+title:something +(hey)'),
-    ('title:something hey', '+title:something +(hey)'),
-    ('hey title:something hey person:john', '+title:something +person:john +(hey hey)'),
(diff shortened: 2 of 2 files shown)


citadel/indico_citadel/util.py
```diff
@@ -67,15 +67,20 @@ def format_query(query, placeholders):
...
-    return ' '.join(p)
+    patt = r'(?:^|\s)({}):([^:"\s]+|"[^"]+")(?:$|\s)'.format('|'.join(placeholders.keys()))
+    idx = 0
+    keys = []
+    for x in re.finditer(patt, query):
+        placeholder = f'+{placeholders[x.group(1)]}:{escape(x.group(2))}'
+        if idx != x.start():
+            keys.append(escape(query[idx:x.start()]))
+        keys.append(placeholder)
+        idx = x.end()
```
ThiefMaster, line 78:
```suggestion
    for match in re.finditer(patt, query):
        placeholder = f'+{placeholders[match.group(1)]}:{escape(match.group(2))}'
        if idx != match.start():
            keys.append(escape(query[idx:match.start()]))
        keys.append(placeholder)
        idx = match.end()
```

citadel/indico_citadel/util.py
```diff
@@ -67,15 +67,20 @@ def format_query(query, placeholders):
...
-    patt = r'({}):([^:"\s]+|"[^"]+")\s*'.format('|'.join(placeholders.keys()))
-    # Extract all placeholders
-    p = [f'+{placeholders[x.group(1)]}:{x.group(2)}'
-         for x in re.finditer(patt, query) if x.group(1) in placeholders]
-    # Escape keyword based arguments
-    query = escape(re.sub(patt, '', query)).strip()
-    if query:
-        p.append(f'+({query})')
-    return ' '.join(p)
+    patt = r'(?:^|\s)({}):([^:"\s]+|"[^"]+")(?:$|\s)'.format('|'.join(placeholders.keys()))
```
ThiefMaster, line 70:
```suggestion
    patt = r'(?:^|\s)({}):([^:"\s]+|"[^"]+")(?:$|\s)'.format('|'.join(map(re.escape, placeholders)))
```

`.keys()` isn't needed and while i would never expect any regex-special chars in a placeholder it feels cleaner to run them through `re.escape()` anyway

ThiefMaster:
Might be better for a separate PR, but when I tried this search query it failed with an error: `room:28/R* NOT room:28/R-015`.. This seems to happen because `foo NOT +bar` is not a valid query. `.replace(' NOT +', ' NOT ')` seems to avoid this, even though it's super hacky ;)

But maybe we could do that (unless you see a better option) and possibly allow the `-` prefix as well?

Like this these queries would work nicely: for example

- `room:28/R* -room:28/R-015` - find things in building 28 but not in 28/R-015
- `room:28/R* NOT room:28/R-015` - same, just different syntax

citadel/indico_citadel/util_test.py
```diff
@@ -12,16 +12,18 @@
...
-     '+title:"my event" +person:john +(ola ola some\\:yes ola)'),
-    ('title:"my title:something"', '+title:"my title:something"'),
-    ('hello', '+(hello)'),
-    ('hey title:something', '+title:something +(hey)'),
-    ('title:something hey', '+title:something +(hey)'),
-    ('hey title:something hey person:john', '+title:something +person:john +(hey hey)'),
-    ('<*\\^()', '+(\\<*\\\\\\^\\(\\))'),
+     '+title:"my event" ola +person:john ola some\\:yes ola'),
+    ('title:"my title:something"', '+title:"my title\\:something"'),
+    ('hello   ', 'hello   '),
```
ThiefMaster, line 17: we could `.strip()` the final string to get rid of this

plourenco:
I actually added the placeholders prefixed `+` to create some backwards compatibility if we test something in between. They aren't supposed to be there, and this "hacky" parsing on our side just shows why it's a bad idea to have two parsers with distinct rules.
I think this should be kept no longer than until we can set the `default_operator=AND`.

citadel/indico_citadel/util_test.py
plourenco: That test was meant to show it keeps the original untouched, but I don't see why we shouldn't strip it

PR #130, opened 2021-05-17
Citadel: Flatten the most common aggregations 

plourenco:


citadel/indico_citadel/util.py
```diff
@@ -104,7 +104,7 @@ def format_filters(params, filters, range_filters):
     :param params: The filter query params
     :param filters: The filter whitelist
     :param range_filters: The range filter whitelist
-    :returns: filters, extracted placeholders
+    :returns filters, extracted placeholders
```
ThiefMaster, line 107: I don't think this is correct; sphinx uses `:returns: whatever` and `:return: whatever` (AFAIK we use the second version more frequently)

citadel/indico_citadel/util.py
```diff
@@ -136,3 +136,36 @@ def remove_none_entries(obj):
...
+
+def format_aggregations(aggregations, filters):
+    """Format aggregations into a bucket dictionary.
+
+    Besides transforming each aggregation, ensures each bucket key
+    contains the most common result as defined from Elastic Search.
+
+    :param aggregations: The raw aggregation object
+    :param filters: The filter whitelist
+    :returns key: {label, buckets: [{key, count}]}
```
ThiefMaster, line 149:
```suggestion
    :return: ``key: {label, buckets: [{key, count}]}`` dict
```

citadel/indico_citadel/util.py
```diff
@@ -136,3 +136,36 @@ def remove_none_entries(obj):
...
+                'count': bucket['docCount']
+            } for bucket in value['buckets']]
+        }
+        for key, value in _flatten(aggregations)
+        if key in filters
+    }
+
+
+def _flatten(obj, target_key='buckets', parent_key=''):
+    if type(obj) is not dict:
```
ThiefMaster, line 166:
```suggestion
    if not isinstance(obj, dict):
```

citadel/indico_citadel/util.py
plourenco: Thanks, we had different formats and I had the impression you would point it out if I didn't choose the right one 🙂

PR #138, opened 2021-05-28
Citadel: Move site out of _data

ThiefMaster:


ThiefMaster:
After discussing with @plourenco and @carantunes this is on hold since it can't be easily done with the existing data we already have on ES.

If there's ever a way to make such changes on Citadel without having to do a full re-export from Indico we may want to reconsider this.

PR #139, opened 2021-06-02
Add allow_admin parameter

javfg:
Related to https://github.com/indico/indico/issues/4912.

### DIFF
--- a/citadel/indico_citadel/search.py
+++ b/citadel/indico_citadel/search.py
@@ -64,7 +64,7 @@ class CitadelProvider(IndicoSearchProvider):
             **filter_query
         }
         # Filter by the objects that can be viewed by users/groups in the `access` argument
-        if access := get_user_access(user) if not admin_override_enabled else ['IndicoAdmin']:
+        if access := get_user_access(user, admin_override_enabled):
             access_string = ','.join(access)
             if len(access_string) > 1024:
                 access_string_gz = base64.b64encode(zlib.compress(access_string.encode(), level=9))
--- a/citadel/indico_citadel/search.py
+++ b/citadel/indico_citadel/search.py
@@ -41,7 +41,7 @@ class CitadelProvider(IndicoSearchProvider):
         return bool(CitadelPlugin.settings.get('search_backend_url') and
                     CitadelPlugin.settings.get('search_backend_token'))
 
-    def search(self, query, user=None, page=1, object_types=(), **params):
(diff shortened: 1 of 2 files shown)


citadel/indico_citadel/search.py
```diff
@@ -41,7 +41,7 @@ def active(cls):
         return bool(CitadelPlugin.settings.get('search_backend_url') and
                     CitadelPlugin.settings.get('search_backend_token'))
 
-    def search(self, query, user=None, page=1, object_types=(), **params):
+    def search(self, query, user=None, page=1, object_types=(), allow_admin=False, **params):
```
ThiefMaster, line 44: I'd still prefer `admin_override_enabled` here. `allow_admin` is usually used to indicate whether to take *potential* admin access into account. but if this is set to true here, it already means that there's a user using their admin access.

citadel/indico_citadel/search.py
```diff
@@ -64,6 +64,8 @@ def search(self, query, user=None, page=1, object_types=(), **params):
         }
         # Filter by the objects that can be viewed by users/groups in the `access` argument
         if access := get_user_access(user):
+            if allow_admin:
+                access.append('IndicoAdmin')
```
ThiefMaster, line 68: We could just use `access = ['IndicoAdmin']` and skip getting the user-specific access altogether

javfg: `IndicoAdmin` would be above any others privileges the user can have, right?

ThiefMaster: yes, it's set on all records, so passing it makes passing the user-specific data unnecessary

citadel/indico_citadel/search.py
```diff
@@ -63,7 +64,7 @@ def search(self, query, user=None, page=1, object_types=(), **params):
             **filter_query
         }
         # Filter by the objects that can be viewed by users/groups in the `access` argument
-        if access := get_user_access(user):
+        if access := get_user_access(user) if not admin_override_enabled else ['IndicoAdmin']:
```
ThiefMaster, line 67: I'm not sure I like the way this line looks like now :D 

I'd either move it to a separate statement (instead of inlining it with `:=`), or pass `admin_override_enabled` as an arg to `get_user_access` and check it in here (could even include the `user.is_admin`  check there just to be completely on the safe side)

javfg: Yes, I was feeling the same. I initially made it into a different line without `:=`. I think the second option is better.

PR #154, opened 2021-07-29
Add SIXPay/Saferpay payment plugin

ThiefMaster:
As discussed via chat and email, we are taking over the third-party [sixpay plugin](https://github.com/maxfischer2781/indico_sixpay) as it was no longer maintained upstream and needed to be updated for Indico 3.0 and the latest Saferpay API.

To be more in line with Indico and the other plugins, we're also re-licensing it from GPL to MIT; this was already approved by all the contributors (@maxfischer2781, @cfgmr, @martinclaus) that made relevant contributions to the original codebase, but to have a paper trail here, I'd appreciate if you could leave a comment here confirming this.

Please also let me know if you want any changes to the way you are credited in the file headers and/or setup.cfg author metadata. For example, if you made the original contributions as part of your organization and would like it mentioned, that'd be perfectly fine of course.


maxfischer2781:
Thanks for taking this on! 👍 

- I'm fine with the change to MIT license.
- I'm fine with being mentioned by name only.

martinclaus:
Big thanks also from my side! 👍
I agree to be mentioned by name only and to change the license to MIT.

cfgmr:
Thank you @ThiefMaster!  😃 
I also agree to the proposed licence change and by being mentioned by name only.

PR #156, opened 2021-08-13
Fixes for unlisted events on livesync plugin

javfg:
This PR adds `None` category checks for the livesync plugin.

### DIFF
--- a/livesync/indico_livesync/handler.py
+++ b/livesync/indico_livesync/handler.py
@@ -91,13 +91,17 @@ def _is_category_excluded(category):
 
 
 def _moved(obj, old_parent, **kwargs):
-    _register_change(obj, ChangeType.moved)
+    # if an unlisted event is moved, it triggers a creation instead
+    if isinstance(obj, Event) and old_parent is None:
+        _register_change(obj, ChangeType.created)
+    else:
+        _register_change(obj, ChangeType.moved)
 
-    new_category = obj if isinstance(obj, Category) else obj.category
-    old_excluded = _is_category_excluded(old_parent) if old_parent else False
-    new_excluded = _is_category_excluded(new_category)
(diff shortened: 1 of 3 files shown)


ThiefMaster:
It looks like we're still inserting some stuff in the queue when creating unlisted events.

livesync/indico_livesync/handler.py
```diff
@@ -111,6 +115,9 @@ def _moved(obj, old_parent, **kwargs):
 def _created(obj, **kwargs):
     if not isinstance(obj, (Event, EventNote, Attachment, Contribution, SubContribution)):
         raise TypeError(f'Unexpected object: {type(obj).__name__}')
+    # we don't care about unlisted events
+    if isinstance(obj, Event) and not obj.category:
```
ThiefMaster, line 119: This is not enough; you're still creating queue entries when adding contributions etc. within an unlisted event.

livesync/indico_livesync/models/queue.py
```diff
@@ -286,11 +286,12 @@ def create(cls, changes, ref, excluded_categories=set()):
                     return
             else:
                 event = obj.folder.event if isinstance(obj, Attachment) else obj.event
-                if event.category not in g.setdefault('livesync_excluded_categories_checked', {}):
-                    g.livesync_excluded_categories_checked[event.category] = \
-                        excluded_categories & set(event.category_chain)
-                if g.livesync_excluded_categories_checked[event.category]:
-                    return
+                if event.category:
```
ThiefMaster, line 289: Maybe we can skip unlisted events altogether in here? Because this code is known to already work with events in blacklisted categories, and in the end being unlisted is just another reason why an event or anything inside should not be added to the livesync queue.

PR #157, opened 2021-10-13
Zoom plugin: Inject bundle in conference web page

javfg:
This PR injects the js for the zoom plugin into the conference web for https://github.com/indico/indico/pull/5124.

### DIFF
--- a/vc_zoom/indico_vc_zoom/plugin.py
+++ b/vc_zoom/indico_vc_zoom/plugin.py
@@ -19,7 +19,7 @@ from indico.core import signals
 from indico.core.auth import multipass
 from indico.core.errors import UserValueError
 from indico.core.plugins import IndicoPlugin, render_plugin_template, url_for_plugin
-from indico.modules.events.views import WPSimpleEventDisplay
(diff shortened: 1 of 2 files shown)


vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -160,6 +160,7 @@ def init(self):
         self.inject_bundle('main.js', WPSimpleEventDisplay)
         self.inject_bundle('main.js', WPVCEventPage)
         self.inject_bundle('main.js', WPVCManageEvent)
+        self.inject_bundle('main.js', WPConferenceDisplayBase)
```
ThiefMaster, line 163: Do you remember if there was a reason to use the base class instead of `WPConferenceDisplay`? By default bundles are also injected in subclasses, which results in the same bundle being included twice on the actual conference videoconference page.

I think regardless of that we should fix that in the core, ie. deduplicate injected bundles or even do it like we do for core bundles in `_resolve_bundles`  (throwing an exception while in debug mode)

javfg: I've changed it to `WPConferenceDisplay` and I think there is no problem. I'll open another PR.

PR #171, opened 2022-02-23
Create a task to reschedule zoom meetings

### LINKED ISSUE #170: Zoom plugin: Handle meeting expirations
Zoom meeting expirations are not handled by the webhook event `meeting.deleted`.
To prevent meetings from expiring unexpectedly, we could periodically pool these for events to happen in the future.

plourenco:
These changes help the zoom meetings cope with event (or contribution, session) `start_dt` changes which could potentially lead to expired meetings for postponed events. The logic is the following:
- When an event is anticipated, meetings remain untouched as they only expire on a future date.
- When an event is postponed, we need to check if the new date is close to the video-conference room expiration, by comparing it with **the original** scheduled date.

For point number 2, I've considered that all meetings part of an event should always be scheduled for the exact same date. Using one of them, we query zoom for its original scheduled date. To conclude, we check if the new event date is 30 days after the former ensuring postponing multiple times won't lead to unnecessary updates.

Might close #170

### DIFF
--- a/vc_zoom/indico_vc_zoom/plugin.py
+++ b/vc_zoom/indico_vc_zoom/plugin.py
@@ -528,9 +528,7 @@ class ZoomPlugin(VCPluginMixin, IndicoPlugin):
 
         @after_this_request
         def _launch_task(response):
-            prev_dt = obj.start_dt
-            refresh_meetings.delay(zoom_rooms, obj.start_dt, log_entry, int(obj.duration.seconds / 60),
-                                   lambda: obj.start_dt == prev_dt)
+            refresh_meetings.delay(zoom_rooms, obj, log_entry)
             return response
 
     def _render_vc_room_labels(self, event, vc_room, **kwargs):
--- a/vc_zoom/indico_vc_zoom/task.py
+++ b/vc_zoom/indico_vc_zoom/task.py
@@ -18,14 +18,12 @@ def update_state_log(log_entry, failed):
 
 
 @celery.task(plugin='vc_zoom', autoretry_for=(HTTPError,))
-def refresh_meetings(vc_rooms, target_dt, log_entry=None, duration=None, condition=None):
+def refresh_meetings(vc_rooms, obj, log_entry=None):
     from indico_vc_zoom.api import ZoomIndicoClient
-    if condition and not condition():
-        return
     client = ZoomIndicoClient()
     failed = False
-    payload = {'start_time': target_dt}
-    if duration:
+    payload = {'start_time': obj.start_dt}
+    if duration := int(obj.duration.total_seconds() / 60):
         payload['duration'] = duration
     try:
         for vc_room in vc_rooms:
--- a/vc_zoom/tests/task_test.py
+++ b/vc_zoom/tests/task_test.py
@@ -5,7 +5,7 @@
 # them and/or modify them under the terms of the MIT License;
 # see the LICENSE file for more details.
 
-from datetime import datetime, timedelta
+from datetime import timedelta
 
 import pytest
 from requests.exceptions import HTTPError
@@ -13,6 +13,7 @@ from requests.exceptions import HTTPError
 from indico.core.db.sqlalchemy.util.session import no_autoflush
 from indico.modules.logs import EventLogRealm, LogKind
 from indico.modules.vc.models.vc_rooms import VCRoom, VCRoomEventAssociation, VCRoomLinkType, VCRoomStatus
+from indico.util.date_time import now_utc
 
 from indico_vc_zoom.util import ZoomMeetingType
 
@@ -35,15 +36,17 @@ def dummy_vc_room(db, dummy_event, dummy_user):
     return room
 
 
-def test_meeting_reschedule(app, db, dummy_event, dummy_user, dummy_vc_room, zoom_client):
+@pytest.mark.usefixtures('dummy_vc_room')
+def test_meeting_reschedule(dummy_event, zoom_client):
     from indico_vc_zoom.task import refresh_meetings
-    meeting_time = datetime.now() - timedelta(days=30)
+    meeting_time = now_utc() - timedelta(days=30)
     log_entry = dummy_event.log(EventLogRealm.event, LogKind.change, 'Test', 'Test', data={'State': 'pending'})
     dummy_event.start_dt = meeting_time + timedelta(days=5)
-    refresh_meetings([room.vc_room for room in dummy_event.vc_room_associations], dummy_event.start_dt, log_entry)
+    refresh_meetings([room.vc_room for room in dummy_event.vc_room_associations], dummy_event, log_entry)
     assert zoom_client.update_meeting.call_count == 1
     assert log_entry.data['State'] == 'succeeded'
     zoom_client.update_meeting.side_effect = HTTPError()
-    refresh_meetings([room.vc_room for room in dummy_event.vc_room_associations], dummy_event.start_dt, log_entry)
+    with pytest.raises(HTTPError):
+        refresh_meetings([room.vc_room for room in dummy_event.vc_room_associations], dummy_event, log_entry)
     assert zoom_client.update_meeting.call_count == 2
     assert log_entry.data['State'] == 'failed'


vc_zoom/indico_vc_zoom/task.py
```diff
@@ -0,0 +1,41 @@
...
+def check_meetings(target, changes):
+    if 'start_dt' not in changes:
+        return
+    prev_date, new_date = changes.get('start_dt')
+    if new_date < prev_date:
+        return
+    refresh_meetings.delay(target.vc_room_associations, target.start_dt)
+
+
+@celery.task(name='reschedule_meeting', plugin='vc_zoom')
```
ThiefMaster, line 28: IIRC we usually do not specify a task name in plugins but rather keep the auto-generated one

vc_zoom/indico_vc_zoom/task.py
```diff
@@ -0,0 +1,41 @@
...
+
+@signals.event.updated.connect
+@signals.event.contribution_updated.connect
+@signals.event.session_block_updated.connect
+def check_meetings(target, changes):
+    if 'start_dt' not in changes:
+        return
+    prev_date, new_date = changes.get('start_dt')
+    if new_date < prev_date:
+        return
```
ThiefMaster, line 24: Shouldn't going back in time, as long as it's still in the future, be handled as well?

vc_zoom/indico_vc_zoom/task.py
```diff
@@ -0,0 +1,41 @@
+# This file is part of the Indico plugins.
+# Copyright (C) 2002 - 2021 CERN
```
ThiefMaster, line 2: 2022

vc_zoom/indico_vc_zoom/task.py
```diff
@@ -0,0 +1,41 @@
...
+                and room.data['meeting_type'] == ZoomMeetingType.scheduled_meeting]
+    if not vc_rooms:
+        return
+    meeting = client.get_meeting(vc_rooms[0].data['zoom_id'])
+    start_dt = datetime.datetime.fromisoformat(meeting['start_time'].replace("Z", "+00:00"))
+    if target_dt < start_dt.replace(tzinfo=None) + datetime.timedelta(days=30):
+        return
+    for vc_room in vc_rooms:
+        serialized_date = target_dt.replace(microsecond=0).isoformat() + 'Z'
+        client.update_meeting(vc_room.data['zoom_id'], {'start_time': serialized_date})
```
ThiefMaster, line 41: I think this might fail if the zoom meeting is already recurring

vc_zoom/indico_vc_zoom/task.py
plourenco: At first sight, I don't see any case. Because meetings should always be aligned with the event date, if we're going to the past, the expiration should always be in the future (and never crossed).

vc_zoom/indico_vc_zoom/task.py
plourenco: I see... but shouldn't the previous filter by `ZoomMeetingType.scheduled_meeting` prevent that?

vc_zoom/indico_vc_zoom/task.py
ThiefMaster: Sure, this isn't so much about expiry but about keeping a scheduled Zoom meeting in sync with the indico event (the zoom client reminds you a few minutes before one of your meetings starts)

vc_zoom/indico_vc_zoom/task.py
ThiefMaster: ah yes probably...

vc_zoom/indico_vc_zoom/task.py
plourenco: Ah well, we can do - that's much simpler. I assumed we only wanted to update when strictly necessary.

vc_zoom/indico_vc_zoom/task.py
plourenco: Just to add that I've been reviewing this with real examples and even though only scheduled meetings should contain a fixed `start_time`, there were no errors whatsoever while sending them to recurring too (type 3), which seems to be ignored.
To avoid querying for the type, I left it as-is.

vc_zoom/indico_vc_zoom/__init__.py
```diff
@@ -5,7 +5,27 @@
...
+from indico.core.db import db
+from indico.modules.logs import EventLogRealm, LogKind
 from indico.util.i18n import make_bound_gettext
+from indico_vc_zoom.task import refresh_meetings
 
 
 _ = make_bound_gettext('vc_zoom')
+
+
+@signals.event.updated.connect
```
plourenco, line 18: Is there any difference between the decorator directly or connecting through the IndicoPlugin class?

ThiefMaster: If you use the decorator, it will run at import time; if you call it in the plugin's init method it will only run when the plugin is actually loaded

plourenco: Yes, I moved into this file to cope with that - is there anything against having it at import time?

ThiefMaster: I would put stuff in the plugin like we do in the other plugins as well; usually `__init__.py` of plugins doesn't contain anything but the i18n binding and (if needed) binding the `import_tasks` signal (i don' remember if that was needed or if i just didn't think about putting it in the plugin when adding it for the first time to ap lugin)

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -516,34 +521,23 @@ def _merge_users(self, target, source, **kwargs):
     def get_notification_cc_list(self, action, vc_room, event):
         return {principal_from_identifier(vc_room.data['host']).email}
 
+    def _check_meetings(self, target, changes):
```
ThiefMaster, line 524:
```suggestion
    def _check_meetings(self, target, changes, **kwargs):
```

always accept extra kwargs in signal handlers

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -152,7 +155,9 @@ class ZoomPlugin(VCPluginMixin, IndicoPlugin):
     def init(self):
         super().init()
         self.connect(signals.plugin.cli, self._extend_indico_cli)
-        self.connect(signals.event.times_changed, self._times_changed)
```
ThiefMaster, line 155: I'm not sure if the new signals really handle everything needed... try e.g. moving a contribution to start before the event start time. This triggers an update of the event's start time, but I don't think it triggers `event.updated`.

So maybe better to keep using `times_changed`? AFAIK this one triggers for all time-related updates in events, contributions, etc.

ThiefMaster: My comment was purely technical - we have signals that handle all date/time changes so IMHO using those is better than using generic "something changed" signals.

But yes, if we keep things in sync I'd be exact - after all the Zoom client typically notifies people about upcoming scheduled meetings...

plourenco: I see what you mean, makes sense.

vc_zoom/indico_vc_zoom/task.py
```diff
@@ -0,0 +1,37 @@
...
+from indico.core.celery import celery
+from indico.core.db import db
+
+
+def update_state_log(log_entry, failed):
+    log_entry.data['State'] = 'failed' if failed else 'succeeded'
+    flag_modified(log_entry, 'data')
+
+
+@celery.task(plugin='vc_zoom', max_retries=None)
```
ThiefMaster, line 20: Does this allow unlimited retries? That may be a bit risky as a failing request would keep failing forever even if the zoom meeting got deleted in the meantime

vc_zoom/indico_vc_zoom/task.py
```diff
@@ -0,0 +1,37 @@
...
+def refresh_meetings(vc_rooms, target_dt, log_entry=None):
+    from indico_vc_zoom.api import ZoomIndicoClient
+    client = ZoomIndicoClient()
+    failed = False
+    try:
+        for vc_room in [room for room in vc_rooms if room.type == 'zoom']:
```
ThiefMaster, line 26: You already filter before calling the desk so no need to check again

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -516,34 +519,21 @@ def _merge_users(self, target, source, **kwargs):
     def get_notification_cc_list(self, action, vc_room, event):
         return {principal_from_identifier(vc_room.data['host']).email}
 
+    def _check_meetings(self, sender, obj, **kwargs):
+        zoom_rooms = [room.vc_room for room in obj.vc_room_associations if room.vc_room.type == 'zoom']
+        log_entry = obj.log(EventLogRealm.event, LogKind.change, 'Videoconference', 'Schedule updated', data={
+            'Meetings': ','.join([f'{room.name} (ID: {room.id})' for room in zoom_rooms]),
```
ThiefMaster, line 525:
```suggestion
            'Meetings': ', '.join([f'{room.name} (ID: {room.id})' for room in zoom_rooms]),
```

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -516,34 +519,21 @@ def _merge_users(self, target, source, **kwargs):
...
+            'Date': obj.start_dt.isoformat(),
+            'State': 'pending'
+        })
+        db.session.commit()
```
ThiefMaster, line 529: doing commit here seem rather dangerous! If this is because of the log entry, you could trigger the celery task after the request:

```py
@after_this_request
def _launch_task(response):
    refresh_meetings.delay(zoom_rooms, obj.start_dt, log_entry)
    return response
```

plourenco: Sounds good

vc_zoom/indico_vc_zoom/task.py
plourenco: Yes, just trying to ensure some consistency.

vc_zoom/indico_vc_zoom/task.py
```diff
@@ -0,0 +1,37 @@
...
+            client.update_meeting(vc_room.data['zoom_id'], {'start_time': target_dt})
```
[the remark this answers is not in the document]
plourenco: No, unfortunately, I would have to query for the type.
There were no errors whatsoever while sending them to recurring too (type 3), which seems to be ignored.
To avoid querying for the type, I left it as-is.

vc_zoom/indico_vc_zoom/task.py
plourenco: I have the impression there was a reason, but definitely risky.
I've also updated the retry mechanism.

PR #178, opened 2022-04-26
Previewer_jupyter: Update dependencies

### LINKED ISSUE #176: Jupyter preview: Update nbconvert dependency
The current (very old) version is broken once we update updating Jinja.

```pytb
2022-04-25 11:49:44,074  ERROR    0000000000000000  -       indico.plugins            Could not load plugin previewer_jupyter
Traceback (most recent call last):
  ...
  File "/home/adrian/dev/indico/py3/plugins/public/previewer_jupyter/indico_previewer_jupyter/controllers.py", line 12, in <module>
    from nbconvert.exporters import HTMLExporter
  File "/home/adrian/dev/indico/py3/env/lib/python3.9/site-packages/nbconvert/__init__.py", line 4, in <module>
    from .exporters import *
  File "/home/adrian/dev/indico/py3/env/lib/python3.9/site-packages/nbconvert/exporters/__init__.py", line 3, in <module>
    from .html import HTMLExporter
  File "/home/adrian/dev/indico/py3/env/lib/python3.9/site-packages/nbconvert/exporters/html.py", line 12, in <module>
    from jinja2 import contextfilter
ImportError: cannot import name 'contextfilter' from 'jinja2' (/home/adrian/dev/indico/py3/env/lib/python3.9/site-packages/jinja2/__init__.py)
```

I remember the last time I tried to update (IIRC during early 3.0) the styling was very broken, so there's probably some customization needed to get a similarly "normal" look&feel when previewing a Jupyter notebook within Indico.

plourenco:
Closes #176


previewer_jupyter/indico_previewer_jupyter/templates/ipynb_preview.html
```diff
@@ -1,31 +1,42 @@
 <!DOCTYPE html>
 <html lang="en">
-    <head>
```
ThiefMaster, line 3: I think our templates still use 4 spaces; it's just JS files that use 2 spaces

previewer_jupyter/indico_previewer_jupyter/templates/ipynb_preview.html
```diff
@@ -1,31 +1,42 @@
...
+      src="https://cdn.jsdelivr.net/npm/katex@0.15.3/dist/katex.min.js"
+      integrity="sha384-0fdwu/T/EQMsQlrHCCHoH10pkPLlKA1jL5dFyUOvB3lfeT2540/2g6YgSi2BL14p"
+      crossorigin="anonymous"
+    ></script>
+  </head>
+  <body>
+    {{ html_code | safe }}
+
+    <script type="module" nonce="{{ nonce }}">
+      import renderMathInElement from 'https://cdn.jsdelivr.net/npm/katex@0.15.3/dist/contrib/auto-render.mjs';
```
ThiefMaster, line 24: can this be protected via https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity? i don't like the idea of embedding random files from a CDN with no guarantees that they are never modified.

alternatively just shipping it w/ the plugin would be an option

plourenco: Right, I guess I can add back the integrity...

previewer_jupyter/indico_previewer_jupyter/controllers.py
```diff
@@ -45,9 +47,9 @@ def _process(self):
 
         response = current_app.response_class(html)
         # Use CSP to restrict access to possibly malicious scripts or inline JS
-        csp_header = f"script-src cdn.mathjax.org 'nonce-{nonce}';"
-        response.headers['Content-Security-Policy'] = csp_header
-        response.headers['X-Webkit-CSP'] = csp_header
+        # csp_header = f"script-src cdnjs.cloudflare.com cdn.jsdelivr.net 'nonce-{nonce}';"
+        # response.headers['Content-Security-Policy'] = csp_header
+        # response.headers['X-Webkit-CSP'] = csp_header
```
ThiefMaster, line 52: We should keep the CSP; it's needed to reduce the XSS risk from previewed notebooks

previewer_jupyter/indico_previewer_jupyter/templates/ipynb_preview.html
```diff
@@ -5,27 +5,42 @@
...
+        <script
+            src="https://cdn.jsdelivr.net/npm/katex@0.15.3/dist/katex.min.js"
+            integrity="sha384-0fdwu/T/EQMsQlrHCCHoH10pkPLlKA1jL5dFyUOvB3lfeT2540/2g6YgSi2BL14p"
+            crossorigin="anonymous"
+        ></script>
+        <script
+            src="https://cdn.jsdelivr.net/npm/katex@0.15.3/dist/contrib/auto-render.min.js"
+            integrity="sha384-+XBljXPPiv+OzfbB3cVmLHf4hdUFHlWNZN5spNQ7rmHTXpd7WvJum6fIACpNNfIR"
+            crossorigin="anonymous"
+        ></script>
```
ThiefMaster, line 23: maybe better to just include them locally and not relying on a cdn at all

ThiefMaster (commented):
also bump the plugin version to 3.2-dev and add a changelog entry

ThiefMaster (approved):

PR #183, opened 2022-07-14
Add Google reCAPTCHA plugin

tomasr8:
First time making a plugin so I might've missed something

Depends on indico/indico#5400

- Warns when site key has not been set
- Automatically reloads when the answer does not succeed

image


recaptcha/indico_recaptcha/plugin.py
```diff
@@ -0,0 +1,60 @@
...
+from indico.web.forms.base import IndicoForm
+from indico.core.plugins import IndicoPlugin
+from indico.modules.events.registration.plugins import CaptchaPluginMixin
+from indico.web.views import WPBase
+from indico_recaptcha import _
+
+
+class ReCaptchaSettingsForm(IndicoForm):
+    site_key = StringField('Site key', [DataRequired()], description=_('The site key available in the reCAPTCHA admin dashboard'))
+    secret_key = StringField('Secret key', [DataRequired()], description=_('The secret key available in the reCAPTCHA admin dashboard'))
```
ThiefMaster, line 22: missing i18n for the field title. also, those lines look excessively long...

recaptcha/indico_recaptcha/client/Captcha.jsx
```diff
@@ -0,0 +1,87 @@
...
+    }
+  }, [submitErrors, name, resetCaptcha]);
+
+  return (
+    <Message info style={{marginTop: 25}}>
+      <Message.Header>
+        <Translate>Confirm that you are not a robot</Translate> 🤖
+      </Message.Header>
+      <div>
+        {!siteKey && (
```
ThiefMaster, line 57: You can do `{siteKey ? ... : ...}` instead of the two separate `&&` blocks...

But I'd maybe add a method to the plugin mixin to check whether the captcha is available and fall back to the built-in one if it isn't. Otherwise an admin may forget it and get unusable regforms by simply having the plugin installed.

An "enabled" plugin setting that also disables it would be nice as well - e.g. to easily disable it if recaptcha has a temporary outage.

ThiefMaster (approved):

PR #187, opened 2022-09-15
Disable add files button when not ready

javfg:
This PR disables the 'Add' button in the 'Add files' dialog for the ownCloud plugin while the File-picker is performing async operations.

This fixes the bug caused by a user clicking too fast on "Add" after selecting a file, where he would get notified that there are no files selected.

### DIFF
--- a/owncloud/indico_owncloud/client/index.js
+++ b/owncloud/indico_owncloud/client/index.js
@@ -9,6 +9,10 @@
 window.setupOwncloudFilePickerWidget = ({filepickerUrl, fieldId}) => {
   window.addEventListener('message', message => {
     const iframe = document.querySelector('#owncloud_filepicker-file-picker');
+    const submitButton = document.querySelector(
+      '#attachment-owncloudfilepicker-form input[type=submit]'
+    );
+
     if (
       iframe &&
       message.origin === filepickerUrl &&
@@ -17,6 +21,7 @@ window.setupOwncloudFilePickerWidget = ({filepickerUrl, fieldId}) => {
       message.data.files
     ) {
       document.getElementById(`${fieldId}-files`).value = message.data.files.join('\n');
+      submitButton.disabled = !message.data.ready || !message.data.files.length;
     }
   });
 };
--- a/owncloud/indico_owncloud/templates/owncloud_filepicker_widget.html
+++ b/owncloud/indico_owncloud/templates/owncloud_filepicker_widget.html
@@ -12,6 +12,7 @@
 
 {% block javascript %}
     <script>
+        document.querySelector('#attachment-owncloudfilepicker-form input[type=submit]').disabled = true;
         setupOwncloudFilePickerWidget({
             filepickerUrl: {{ field.filepicker_url | tojson }},
             fieldId: {{ field.id | tojson }},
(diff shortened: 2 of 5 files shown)


owncloud/indico_owncloud/client/index.js
```diff
@@ -9,6 +9,8 @@
 window.setupOwncloudFilePickerWidget = ({filepickerUrl, fieldId}) => {
   window.addEventListener('message', message => {
     const iframe = document.querySelector('#owncloud_filepicker-file-picker');
+    const submitButton = document.querySelectorAll('input[type=submit]')[0];
```
ThiefMaster, line 12:
```suggestion
    const submitButton = document.querySelector('input[type=submit]');
```

However, I believe this is too generic since it can match *any* submit button on that page...

owncloud/indico_owncloud/client/index.js
```diff
@@ -17,6 +19,12 @@ window.setupOwncloudFilePickerWidget = ({filepickerUrl, fieldId}) => {
       message.data.files
     ) {
       document.getElementById(`${fieldId}-files`).value = message.data.files.join('\n');
+
+      if (!message.data.ready || !message.data.files.length) {
+        submitButton.setAttribute('disabled', '');
```
ThiefMaster, line 24: `submitButton.disabled = <bool>;`

owncloud/indico_owncloud/templates/owncloud_filepicker_widget.html
```diff
@@ -12,6 +12,7 @@
 
 {% block javascript %}
     <script>
+        document.querySelectorAll('input[type=submit]')[0].setAttribute('disabled', '');
```
ThiefMaster, line 15: you can use a bool here as well ;)

owncloud/indico_owncloud/client/index.js
javfg: I've seen that if I set it to a bool, it will add `disabled="true"` to the DOM, but isn't just `disabled` more correct?

ThiefMaster: https://stackoverflow.com/a/23701672/298479

Setting the property is more convenient (bool) and controls the attribute internally.

owncloud/indico_owncloud/templates/owncloud_filepicker_widget.html
```diff
@@ -12,6 +12,7 @@
 
 {% block javascript %}
     <script>
+        document.querySelector('input[type=submit]').disabled = true;
```
ThiefMaster, line 15: This is still way too generic ;/ It'll disable submit inputs on the whole page (even though there may not be any others).

```suggestion
        document.querySelector('#attachment-owncloudfilepicker-form input[type=submit]').disabled = true;
```

owncloud/indico_owncloud/client/index.js
```diff
@@ -9,6 +9,8 @@
 window.setupOwncloudFilePickerWidget = ({filepickerUrl, fieldId}) => {
   window.addEventListener('message', message => {
     const iframe = document.querySelector('#owncloud_filepicker-file-picker');
+    const submitButton = document.querySelector('input[type=submit]');
```
ThiefMaster, line 12:
```suggestion
    const submitButton = document.querySelector(
      '#attachment-owncloudfilepicker-form input[type=submit]'
    );
```

PR #188, opened 2022-09-29
Updated dependency to Python 3.10

andriish:
Updated dependency to Python 3.10


ThiefMaster:
Cool, totally forgot about that ;)
Mind doing the same PR for the plugins in https://github.com/indico/indico-plugins-cern as well?

PR #189, opened 2022-09-29
Update dependency on iso4217

andriish:
Update dependency on iso4217


andriish:
Hi @ThiefMaster ,

would you have a look? 

Best regards,

Andrii

ThiefMaster:
I'd still greatly prefer switching to the latest version but not using a less strict version pin. Feel free to patch this when building an RPM though to allow newer versions.

PR #190, opened 2022-09-29
Update dependency on nbconvert

andriish:
Update dependency on nbconvert

### DIFF
--- a/previewer_jupyter/setup.cfg
+++ b/previewer_jupyter/setup.cfg
@@ -21,7 +21,7 @@ include_package_data = true
 python_requires = >=3.9.0, <3.11
 install_requires =
     indico>=3.2
-    nbconvert==6.5.0
+    nbconvert>=6.5.0
 
 [options.entry_points]
 indico.plugins =


previewer_jupyter/setup.cfg
```diff
@@ -21,7 +21,7 @@ include_package_data = true
 python_requires = >=3.9.0, <3.11
 install_requires =
     indico>=3.2
-    nbconvert==6.5.0
+    nbconvert>=6.5.0
```
ThiefMaster, line 24: This accepts new major versions and considering that the last major bump did not Just Work without any changes, I don't like the idea of using an unbounded version like this. I'd rather bump the pin to `nbconvert==7.0.0` and make sure that everything still works as it did before (proper styling etc).

Unfortunately their [changelog](https://nbconvert.readthedocs.io/en/latest/changelog.html#id1) doesn't mention the reason for the major version bump...

andriish: Hi @ThiefMaster ,

would something like 
```
nbconvert>=6.5.0,<7.0.0
```
make more sense?

Pining things up to the patch version is not a good idea.

Andrii
P.S. I don't actually remember the syntax for the version limits.

ThiefMaster: I know that it's usually not recommended, but I believe this recommendation is more about libraries and not applications.

In a library you typically want to give the person installing it the freedom to use a newer version. But in an application you have one set of dependencies that you know is compatible.

Also, not pinning a specific version could be a security problem in case the `nbconvert` pypi account gets compromised at some point - a malicious release would likely just be a patch version bump to get he least amount of scrutiny from people updating their deps (or even none at all due to more loose pins).

andriish: Hi @ThiefMaster ,

OK, I see your point. 
From the point of view of building the RPMs, it is not a problem to insert a one-liner into a spec file that would replace the dependency. And the security problems are mitigated with explicit checks of versions and md5 sums.
So in the end it is just about the policy. If you prefer to strictly pin the versions -- let it be so.
Then it makes sense to close this PR as well as https://github.com/indico/indico-plugins/pull/189.
Will do that.

And purely in the discussion mode:
 - For the security reasons instead of the strict == requirement it could make sense to have something like
```
nbconvert>=6.5.0,<6.5.X
```
where X is a known released version. Or, even a combination, e.g.
```
nbconvert>=6.5.0,<6.5.X || >=7.0.0,<7.0.Y
```


Best regards,

Andrii

PR #192, opened 2022-11-30
Do not break Zoom meetings with registration enabled

### LINKED ISSUE #191: Zoom: Do not break meetings with registration enabled
> I discovered that in certain conditions, assigning an existing Zoom meeting to an Indico event changes that Zoom meeting from scheduled to recurring.
> This by itself is usually not a problem - unless that Zoom meeting required registration. Registration can be required only for scheduled Zoom meetings, and so changing it to a recurring meeting effectively means that registration is not longer required to join that meeting.
> This may be very problematic for Zoom meetings when having obligatory registration was crucial - I've been bitten by exactly this earlier this week :slightly_frowning_face:

A solution would be to check whether the Zoom meeting uses registration and in that case:

- prohibit linking it to another event
- prohibit cloning that event with the VC-cloner enabled

ThiefMaster:
Such meetings will be skipped while cloning now and the user gets an error when trying to assign such a meeting to antoher event.

fixes #191


vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -448,28 +456,42 @@ def clone_room(self, old_event_vc_room, link_object):
...
+            if vc_room.status != VCRoomStatus.deleted:
+                # mark room as deleted
+                vc_room.status = VCRoomStatus.deleted
+                flash(
+                    _('The room "{}" no longer exists in Zoom and was removed from the event').format(vc_room.name),
+                    'warning'
+                )
+            return None
+
+        if vc_room.data.get('registration_required'):
```
ThiefMaster, line 479: test, wth is this new thing on vscode?

PR #197, opened 2023-02-23
ci: bump versions of actions and update depcrecated syntax

olifre:
`set-output` and `save-state` will fail to work in June 2023: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Note: Action versions can also be kept up-to-date via dependapot in case you prefer automation, see e.g.:
https://github.com/unibonn/indico-sso-group-mapping/blob/ed9eb4264988880fa3abe9fc5d48f589637e0fb0/.github/dependabot.yml

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
         shell: bash
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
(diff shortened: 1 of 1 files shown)


ThiefMaster:
Thanks! If this works fine, feel free to send PRs to `indico-plugins-cern` and the main `indico` repos as well ;)

PR #201, opened 2023-04-19
Prometheus: new plugin to expose metrics

pferreir:
A simple plugin which adds a `/metrics` endpoint to Indico, thus allowing it to be queried by Prometheus. Query results can be cached, and the endpoint can be protected using a bearer token.


prometheus/indico_prometheus/blueprint.py
```diff
@@ -0,0 +1,18 @@
...
+# see the LICENSE file for more details.
+
+from indico.core.plugins import IndicoPluginBlueprint
+
+from indico_prometheus.controllers import RHMetrics
+
+
+blueprint = IndicoPluginBlueprint(
+    'prometheus', __name__,
+    url_prefix='/'
```
ThiefMaster, line 15: not needed and having the `/` here would actually build urls with `//` since the prefix is simply prepended to the paths below.

prometheus/indico_prometheus/metrics.py
```diff
@@ -0,0 +1,149 @@
...
+
+from indico_prometheus.queries import get_attachment_query, get_note_query
+
+
+# Check for availability of the LiveSync plugin
+LIVESYNC_AVAILABLE = True
+try:
+    from indico_livesync.models.queue import LiveSyncQueueEntry
+except ImportError:
+    LIVESYNC_AVAILABLE = False
```
ThiefMaster, line 31: This check is not super reliable, since installing `indico-plugins` will install the livesync plugin as well. But unless you can disable a metric at a later point (not during import time) I don't see a better solution.
But in case you *can* disable a metric later, then I'd check if the livesync plugin is actually loaded.

prometheus/indico_prometheus/metrics.py
```diff
@@ -0,0 +1,149 @@
...
+        .filter(Identity.last_login_dt > (now - timedelta(hours=active_user_hours)))
+        .join(Identity).group_by(User).count()
+    )
+    num_categories.set(Category.query.filter(~Category.is_deleted).count())
+
+    num_active_attachment_files.set(get_attachment_query().count())
+    num_attachment_files.set(
+        AttachmentFile
+        .query
+        .join(Attachment, AttachmentFile.attachment_id == Attachment.id)
```
ThiefMaster, line 92: I think think the second argument isn't needed since there's already a relationship between the two models

prometheus/indico_prometheus/metrics.py
```diff
@@ -0,0 +1,149 @@
...
+        .count()
+    )
+
+    attachment_subq = db.aliased(Attachment, get_attachment_query().subquery("attachment"))
```
ThiefMaster, line 96:
```suggestion
    attachment_subq = db.aliased(Attachment, get_attachment_query().subquery('attachment'))
```

(i need to finally run pyquotes to convert the last few double quotes in the plugin repo and then re-enable the flake8 rule for this ;))

prometheus/indico_prometheus/metrics.py
```diff
@@ -0,0 +1,149 @@
...
+        .query
+        .filter(
+            ~Room.is_deleted,
+            ~Reservation.is_rejected,
+            ReservationOccurrence.is_valid,
+            ReservationOccurrence.start_dt < now,
+            ReservationOccurrence.end_dt > now
+        ).join(Reservation)
+         .join(Room)
+         .count()
```
ThiefMaster, line 148: too much indentation i think... probably best to just wrap after the `)` from `.filter()`

prometheus/indico_prometheus/plugin.py
```diff
@@ -0,0 +1,48 @@
...
+
+from indico.core.plugins import IndicoPlugin
+from indico.web.forms.base import IndicoForm
+
+from indico_prometheus import _
+from indico_prometheus.blueprint import blueprint
+
+
+class PluginSettingsForm(IndicoForm):
+    cache_ttl = IntegerField(_('Cache TTL (s)'), [DataRequired()], description=_('TTL for cache (seconds)'))
```
ThiefMaster, line 19: `TimeDeltaField`?

prometheus/setup.cfg
```diff
@@ -0,0 +1,32 @@
+[metadata]
+name = indico-plugin-prometheus
+version = 0.1
```
ThiefMaster, line 3:
```suggestion
version = 3.2-dev
```

prometheus/indico_prometheus/metrics.py
pferreir: Not sure metrics can be disabled, at least I didn't find any way to do it. Maybe I can check the config at load time? The app should be already set up by now, right?

prometheus/indico_prometheus/metrics.py
pferreir: You need black. Really.

prometheus/indico_prometheus/metrics.py
pferreir: There's more than one, so you have to explicitly specify it.

pferreir:
Also added some unit tests.

prometheus/README.md
```diff
@@ -0,0 +1,22 @@
...
+    scheme: https
+    static_configs:
+      - targets:
+        - yourindicoserver.example.com
+    # it is recommended that you set a bearer token in the config
+    authorization:
+        credentials: inds_metrics_xxxxxx
+    # this is only needed in development setups
+    tls_config:
+      insecure_skip_verify: true
```
ThiefMaster, line 21: Maybe better to default to false and mention that in development it may be necessary to set it to true? People tend to copy&paste and not change things that don't absolutely need to be changed :D

prometheus/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+## prometheus.yml
+```yaml
+scrape_configs:
+  - job_name: indico_stats
+    metrics_path: /metrics
+    scheme: https
+    static_configs:
+      - targets:
+        - yourindicoserver.example.com
```
ThiefMaster, line 15: this doesn't need an `https://` prefix?

prometheus/indico_prometheus/controllers.py
```diff
@@ -0,0 +1,55 @@
...
+from indico.web.rh import RH
+
+from indico_prometheus.metrics import update_metrics
+
+
+cache = make_scoped_cache('prometheus_metrics')
+
+
+class RHMetrics(RH):
+    CSRF_ENABLED = False
```
ThiefMaster, line 24: Not strictly needed if the endpoint is only using GET (which is probably the case since the url route only allows GET)

prometheus/indico_prometheus/controllers.py
```diff
@@ -0,0 +1,55 @@
...
+                REGISTRY, accept_header, accept_encoding_header, request.args, False
+            )
+            cache.set('metrics', (status, headers, output), timeout=current_plugin.settings.get('cache_ttl'))
+
+        resp = make_response(output)
+        resp.status = status
+
+        resp.headers['X-Cached'] = 'yes' if cached else 'no'
+        for key, val in headers:
+            resp.headers[key] = val
```
ThiefMaster, line 53: not sure if the headers collection supports it, but if it does something like `resp.headers.extend(headers)` may work

prometheus/indico_prometheus/metrics.py
```diff
@@ -0,0 +1,150 @@
...
+            Reservation.is_pending,
+            ~Reservation.is_archived
+        ).join(Room).count()
+    )
+
+    num_occurrences.set(ReservationOccurrence.query.count())
+    num_valid_occurrences.set(
+        ReservationOccurrence
+        .query
+        .filter(~Room.is_deleted, ~Reservation.is_rejected, ReservationOccurrence.is_valid)
```
ThiefMaster, line 132: I think you also want to exclude cancelled ones here...  or filter by `Reservation.is_accepted` instead?

prometheus/indico_prometheus/metrics.py
```diff
@@ -0,0 +1,150 @@
...
+        .join(Room)
+        .count()
+    )
+
+    num_ongoing_occurrences.set(
+        ReservationOccurrence
+        .query
+        .filter(
+            ~Room.is_deleted,
+            ~Reservation.is_rejected,
```
ThiefMaster, line 143: likewise

prometheus/indico_prometheus/metrics.py
```diff
@@ -0,0 +1,150 @@
...
+)
+size_attachment_files = metrics.Gauge(
+    'indico_size_attachment_files',
+    'Total size of all attachment files, including older versions / deleted (bytes)'
+)
+
+num_notes = metrics.Gauge('indico_num_notes', 'Number of notes')
+
+num_active_rooms = metrics.Gauge('indico_num_active_rooms', 'Number of active rooms')
+num_rooms = metrics.Gauge('indico_num_rooms', 'Number of rooms')
```
ThiefMaster, line 60: "Number of restricted rooms" could also be an interesting metric (number of rooms with `Room.is_protected`)

ThiefMaster: And "Number of rooms requiring manual booking confirmation" (`reservations_need_confirmation`)

prometheus/indico_prometheus/plugin.py
```diff
@@ -0,0 +1,59 @@
...
+        [DataRequired()], description=_('TTL for cache'),
+        units=('seconds', 'minutes', 'hours')
+    )
+    token = StringField(
+        _('Bearer Token'),
+        [Optional()],
+        description=_("Authentication bearer token for Prometheus. Please note that it should be "
+                      "preceded by 'inds_metrics_'")
+    )
+    active_user_hours = IntegerField(
```
ThiefMaster, line 34: also a candidate for `TimeDeltaField`?

prometheus/README.md
pferreir: No, that's how the Prometheus config works. It's all about hostnames.

prometheus/README.md
```diff
@@ -0,0 +1,22 @@
...
+    # it is recommended that you set a bearer token in the config
+    authorization:
+        credentials: inds_metrics_xxxxxx
+    # this is only needed in development setups
+    tls_config:
+      insecure_skip_verify: true
+```
```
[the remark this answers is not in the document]
pferreir: We have changelogs for plugins? Where?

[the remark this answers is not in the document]
pferreir: Ah, the README. Got you.

prometheus/indico_prometheus/metrics.py
pferreir: Got it.

prometheus/indico_prometheus/metrics.py
pferreir: But can a reservation be pending and cancelled at the same time? It's a single state field.

pferreir: Good point.

prometheus/indico_prometheus/metrics.py
pferreir: Yeah.

prometheus/indico_prometheus/plugin.py
pferreir: Yeah, I initially thought "hours" would be enough granularity, but it's true that "days" is also a possibility.

prometheus/setup.cfg
```diff
@@ -0,0 +1,32 @@
...
+    Programming Language :: Python :: 3.10
+
+[options]
+packages = find:
+zip_safe = false
+include_package_data = true
+python_requires = >=3.9.0, <3.11
+install_requires =
+    indico>=3.2
+    prometheus-client
```
ThiefMaster, line 25:
```suggestion
    prometheus-client==0.16.0
```

they have a low release frequency and with 0.x versions i wouldn't be surprised of breaking changes between releases

prometheus/indico_prometheus/plugin.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class PluginSettingsForm(IndicoForm):
+    cache_ttl = TimeDeltaField(
+        _('Cache TTL'),
+        [DataRequired()],
+        description=_('TTL for cache'),
+        units=('seconds', 'minutes', 'hours')
+    )
+    token = StringField(
```
ThiefMaster, line 29: I'd use an `IndicoPasswordField` with `toggle=True` for this, and add validation for the prefix:

```py
    def validate_token(self, field):
        if field.data and not field.data.startswith('inds_metrics_'):
            raise ValidationError(_('...'))
```

pferreir: I thought asking the user to write the prefix would be a bit cumbersome. How do you see that?

ThiefMaster: oh it's handled implicitly. nevermind in that case. in that case the description is kind of confusing though...

prometheus/indico_prometheus/controllers.py
```diff
@@ -0,0 +1,53 @@
...
+cache = make_scoped_cache('prometheus_metrics')
+
+
+@custom_auth
+class RHMetrics(RH):
+    def _check_access(self):
+        token = current_plugin.settings.get('token')
+        if token:
+            if token != request.bearer_token:
+                raise Unauthorized
```
ThiefMaster, line 29:
```suggestion
        if token and token != request.bearer_token:
            raise Unauthorized
```

prometheus/indico_prometheus/metrics.py
ThiefMaster: I think accessing the config at import time is dangerous. It MAY work when the plugin is being loaded, but any "normal" import outside an app context would break. Probably better to just keep it as-is...

pferreir:
Added "enable" toggle field in settings.

ThiefMaster:
also pushed to pypi :)

PR #222, opened 2024-02-12
Citadel: Implement warning threshold

GovernmentPlates:
This PR implements a warning message that is displayed to category admins when the total count of events in said category goes over a configurable threshold. The warning message will recommend admins to use groups instead.

TODO:
- [x] Configurable event warning threshold (plugin settings)
- [x] Better/More informative message (with styling)
- [x] Cleanup/Refactor

Preview:
image

### DIFF
--- a/citadel/indico_citadel/plugin.py
+++ b/citadel/indico_citadel/plugin.py
@@ -5,6 +5,7 @@
 # them and/or modify them under the terms of the MIT License;
 # see the LICENSE file for more details.
 
+from flask_pluginengine.plugin import render_plugin_template
 from wtforms.fields import BooleanField, IntegerField, URLField
 from wtforms.validators import URL, DataRequired, NumberRange
 
@@ -18,7 +19,6 @@ from indico_citadel import _
 from indico_citadel.backend import LiveSyncCitadelBackend
 from indico_citadel.cli import cli
 from indico_livesync import LiveSyncPluginBase
-from indico_citadel.util import check_event_categories
 
 
 class CitadelSettingsForm(IndicoForm):
@@ -90,7 +90,7 @@ class CitadelPlugin(LiveSyncPluginBase):
         super().init()
         self.connect(signals.core.get_search_providers, self.get_search_providers)
         self.connect(signals.plugin.cli, self._extend_indico_cli)
-        self.template_hook('category-protection-page', check_event_categories)
+        self.template_hook('category-protection-page', self.check_event_categories)
 
     def get_search_providers(self, sender, **kwargs):
         from indico_citadel.search import CitadelProvider
@@ -98,3 +98,10 @@ class CitadelPlugin(LiveSyncPluginBase):
 
     def _extend_indico_cli(self, sender, **kwargs):
         return cli
+
+    def check_event_categories(self, category):
+        threshold = self.settings.get('large_category_warning_threshold')
+        num_events = category.deep_events_count
+
+        if threshold and num_events > threshold:
+            return render_plugin_template('event_category_warning.html')
--- a/citadel/indico_citadel/plugin.py
+++ b/citadel/indico_citadel/plugin.py
@@ -102,7 +102,5 @@ class CitadelPlugin(LiveSyncPluginBase):
 
     def _check_event_categories(self, category):
         threshold = self.settings.get('large_category_warning_threshold')
-        num_events = category.deep_events_count
-
-        if threshold and num_events > threshold:
+        if threshold and category.deep_events_count > threshold:
             return render_plugin_template('event_category_warning.html')
--- a/citadel/indico_citadel/plugin.py
+++ b/citadel/indico_citadel/plugin.py
@@ -18,6 +18,7 @@ from indico_citadel import _
 from indico_citadel.backend import LiveSyncCitadelBackend
 from indico_citadel.cli import cli
 from indico_livesync import LiveSyncPluginBase
+from indico_citadel.util import check_event_categories
 
 
 class CitadelSettingsForm(IndicoForm):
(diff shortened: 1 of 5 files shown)


citadel/indico_citadel/plugin.py
```diff
@@ -75,13 +82,15 @@ class CitadelPlugin(LiveSyncPluginBase):
...
         'disable_search': False,
+        'large_category_warning_threshold': 0,
     }
     backend_classes = {'citadel': LiveSyncCitadelBackend}
 
     def init(self):
         super().init()
         self.connect(signals.core.get_search_providers, self.get_search_providers)
         self.connect(signals.plugin.cli, self._extend_indico_cli)
+        self.template_hook('event-category-threshold-alert', check_event_categories)
```
ThiefMaster, line 93: This name is way too specific - template hooks never describe a particular thing they are used for but something generic, like `category-protection-page`

citadel/indico_citadel/util.py
```diff
@@ -215,3 +215,14 @@ def get_user_access(user, admin_override_enabled=False):
...
+def check_event_categories(category):
+        from indico_citadel.plugin import CitadelPlugin
+        from indico.modules.events import Event
+
+        threshold = CitadelPlugin.settings.get('large_category_warning_threshold')
+        num_events = (Event.query.with_parent(category).count())
+
+        if threshold and num_events > threshold:
+            # TODO: Translate + Better message (e.g. recommend 'groups' instead of 'categories')
+            return f"The category '{category.title}' contains {num_events} events, which may cause performance issues. "
```
ThiefMaster, line 228: Even when just using the signal you can `render_plugin_template` for this. That way you can also nicely wrap it in the `message_box` macro.

citadel/indico_citadel/templates/event_category_warning.html
```diff
@@ -0,0 +1,9 @@
+{% from 'message_box.html' import message_box %}
+
+{% call message_box('warning', large_icon=true, fixed_width=true) %}
+    {% trans category_title=category.title, max_events_threshold=threshold -%}
+        <b>The '{{ category_title }}' category contains over {{ max_events_threshold }} events - this may cause performance issues.</b>
+        <br/>
+        Please consider using groups instead of individual users to avoid having to re-send large amounts of changes to users individually.
```
ThiefMaster, line 7: The category title is already shown on the page, no need to repeat it.

I would show something simple like this:

> **This category contains a large number of events.**
> Please consider using groups instead of individual users when granting access or management privileges since any change to the access control list requires re-synchronizing all events with Indico's search engine.

(not 100% sure about the second sentence though)

citadel/indico_citadel/plugin.py
```diff
@@ -75,13 +82,15 @@ class CitadelPlugin(LiveSyncPluginBase):
...
         'disable_search': False,
+        'large_category_warning_threshold': 0,
     }
     backend_classes = {'citadel': LiveSyncCitadelBackend}
 
     def init(self):
         super().init()
         self.connect(signals.core.get_search_providers, self.get_search_providers)
         self.connect(signals.plugin.cli, self._extend_indico_cli)
+        self.template_hook('category-protection-page', check_event_categories)
```
ThiefMaster, line 93: the callback can be a method in here, no need to have it outside this class

citadel/indico_citadel/templates/event_category_warning.html
GovernmentPlates: I think this sounds pretty concise (and it can be changed later on).

citadel/indico_citadel/plugin.py
```diff
@@ -51,6 +52,12 @@ class CitadelSettingsForm(IndicoForm):
                                                 'is used, the internal Indico search interface will be used. This may '
                                                 'be useful when you are still running a larger initial export and do '
                                                 'not want people to get incomplete search results during that time.'))
+    large_category_warning_threshold = IntegerField(_('Large Category Warning Threshold'),
+                                                    [NumberRange(min=0)],
+                                                    description=_('This shows a warning message to admins when '
```
ThiefMaster, line 57: "admins" is misleading. What about something like this?

> Displays a warning to category managers when changing the ACL of big categories that would result in sending a large amount of data to the Citadel server.

FWIW this is not about incomplete search results but rather about putting high load on the infrastructure.

citadel/indico_citadel/plugin.py
```diff
@@ -75,17 +82,26 @@ class CitadelPlugin(LiveSyncPluginBase):
...
+        self.template_hook('category-protection-page', self.check_event_categories)
 
     def get_search_providers(self, sender, **kwargs):
         from indico_citadel.search import CitadelProvider
         return CitadelProvider
 
     def _extend_indico_cli(self, sender, **kwargs):
         return cli
+
+    def check_event_categories(self, category):
```
ThiefMaster, line 102: I usually start those methods with an underscore since they are internal to the plugin.

citadel/indico_citadel/templates/event_category_warning.html
```diff
@@ -0,0 +1,10 @@
+{% from 'message_box.html' import message_box %}
+
+{% call message_box('warning', large_icon=true, fixed_width=true) %}
+    {% trans -%}
+        <b>This category contains a large number of events.</b>
+        <br/>
```
ThiefMaster, line 6: We don't use XHTML -> `<br>`

citadel/indico_citadel/templates/event_category_warning.html
```diff
@@ -0,0 +1,10 @@
+{% from 'message_box.html' import message_box %}
+
+{% call message_box('warning', large_icon=true, fixed_width=true) %}
+    {% trans -%}
+        <b>This category contains a large number of events.</b>
```
ThiefMaster, line 5: `<strong>...</strong>`

citadel/indico_citadel/templates/event_category_warning.html
ThiefMaster: I would split the two parts into separate trans blocks. What way all the HTML can stay outside which is generally nicer :)

PR #232, opened 2024-06-06
Fix asyncronous wait for paypal confirmation and added the custom pay…

rabser:
Added the paypal fees to be paid by participants, so the effective costs of the conference are not substracted by paypal commission (configurable). Also it's better managed the asynchronous confirmation from paypal IPN and now the return page for the merchant it's showed always as paid.

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -120,13 +120,13 @@ jobs:
           path: dist
       # Try uploading to Test PyPI first, in case something fails.
       - name: Publish to Test PyPI 🧪
-        uses: pypa/gh-action-pypi-publish@v1.10.0
+        uses: pypa/gh-action-pypi-publish@v1.10.1
         with:
           repository-url: https://test.pypi.org/legacy/
           packages-dir: dist/
           skip-existing: true
       - name: Publish to PyPI 🚀
-        uses: pypa/gh-action-pypi-publish@v1.10.0
+        uses: pypa/gh-action-pypi-publish@v1.10.1
         with:
           packages-dir: dist/
           skip-existing: true


payment_paypal/indico_payment_paypal/controllers.py
```diff
@@ -84,14 +88,22 @@ def _verify_business(self):
         return False
 
     def _verify_amount(self):
-        expected_amount = self.registration.price
+        paypal_fixed_fee = float(current_plugin.event_settings.get(self.registration.registration_form.event, 'paypal_fixed_fee'))
+        paypal_percent_fee = float(re.findall(r'^([0-9]+(\.[0-9]+)?)(%)?', current_plugin.event_settings.get(self.registration.registration_form.event, 'paypal_percent_fee'))[0][0])
```
duartegalvao, line 92: regex is not really the most appropriate way to parse numbers. and ideally this should be validated when being input into the settings..

payment_paypal/indico_payment_paypal/controllers.py
```diff
@@ -84,14 +88,22 @@ def _verify_business(self):
...
+
+        expected_amount = float(self.registration.price)
+
+        current_plugin.logger.info("Checking PayPal payment applying fees to expected amount: %s fixed: %s percent: %s", expected_amount, paypal_fixed_fee, paypal_percent_fee)
+
+        expected_amount_with_paypal_fees = round (( expected_amount + paypal_fixed_fee ) / ( 1 - ( paypal_percent_fee / 100 )), 2)
```
duartegalvao, line 98: please keep lines of code up to 121 characters maximum for readability, and remove unnecessary blank lines

payment_paypal/indico_payment_paypal/controllers.py
```diff
@@ -107,13 +119,28 @@ class RHPaypalSuccess(RHPaypalIPN):
...
+        while ((not registration.is_paid) and (sleeps < 30)):
+            current_plugin.logger.warning("Waiting for transaction registration %s",sleeps)
+            time.sleep(1)
+            sleeps=sleeps+1
+            db.session.expire(registration)
+            registration = Registration.query.filter_by(uuid=self.token).first()
+
         flash(_('Your payment request has been processed.'), 'success')
-        return redirect(url_for('event_registration.display_regform', self.registration.locator.registrant))
+        redir_url=url_for('event_registration.display_regform', self.registration.locator.registrant)+"&utime="+str(time.time())
```
duartegalvao, line 134: This is not the appropriate way to pass parameters to a URL. you can simply pass them as named arguments of `url_for`:
```suggestion
        redir_url=url_for('event_registration.display_regform', self.registration.locator.registrant, utime=str(time.time()))
```

payment_paypal/indico_payment_paypal/controllers.py
```diff
@@ -107,13 +119,28 @@ class RHPaypalSuccess(RHPaypalIPN):
...
+        current_plugin.logger.warning("Payment success, now redirect to %s",redir_url)
```
duartegalvao, line 135: No need to print URLs to the logger, it will just create noise (and again, this should be a `debug` message and not `warning`)

payment_paypal/indico_payment_paypal/controllers.py
```diff
@@ -107,13 +119,28 @@ class RHPaypalSuccess(RHPaypalIPN):
...
 
 
 class RHPaypalCancel(RHPaypalIPN):
     """Cancellation message"""
 
     def _process(self):
         flash(_('You cancelled the payment process.'), 'info')
-        return redirect(url_for('event_registration.display_regform', self.registration.locator.registrant))
+        redir_url=url_for('event_registration.display_regform', self.registration.locator.registrant)+"&utime="+str(time.time())
+        current_plugin.logger.warning("Payment cancelled, now redirect to %s",redir_url)
```
duartegalvao, line 145: same here

payment_paypal/indico_payment_paypal/plugin.py
```diff
@@ -31,6 +34,12 @@ class EventSettingsForm(PaymentEventSettingsFormBase):
     business = StringField(_('Business'), [UsedIf(lambda form, _: form.enabled.data), DataRequired(),
                                            validate_business],
                            description=_('The PayPal ID or email address associated with a PayPal account.'))
+    paypal_fixed_fee = StringField(_('PayPal Fixed Transaction Fee'), [UsedIf(lambda form, _: form.enabled.data), DataRequired(),
+                                           validate_paypal_fixed_fee],
+                           description=_('The PayPal transaction fixed fee.'))
+    paypal_percent_fee = StringField(_('PayPal Percentage Transaction Fee'), [UsedIf(lambda form, _: form.enabled.data), DataRequired(),
+                                           validate_paypal_percent_fee],
+                           description=_('The PayPal transaction percentage fee.'))
```
duartegalvao, line 42: it doesn't make much sense to use string fields for this, WTForms has a `DecimalField` that would be ideal for these values, and if you use the validator `NumberRange(min=0, max=999999999.99)` it does all of the validation without the need for regex

payment_paypal/indico_payment_paypal/plugin.py
```diff
@@ -46,7 +55,9 @@ class PaypalPaymentPlugin(PaymentPluginMixin, IndicoPlugin):
                         'business': ''}
     default_event_settings = {'enabled': False,
                               'method_name': None,
-                              'business': None}
+                              'business': None,
+                              'paypal_fixed_fee': "0.0",
+                              'paypal_percent_fee': "0%"}
```
duartegalvao, line 60: again here, these should be decimals
```suggestion
                              'paypal_fixed_fee': Decimal('0.0'),
                              'paypal_percent_fee': Decimal('0.0')}
```

payment_paypal/indico_payment_paypal/util.py
```diff
@@ -21,3 +21,19 @@ def validate_business(form, field):
...
+    if not re.match(r'^[0-9]+(\.[0-9]+)?$', field.data):
+        raise ValidationError(_('Invalid PayPal Fixed Fee value'))
+
+def validate_paypal_percent_fee(form, field):
+    """Validates a PayPal percent fee value.
+
+    It can either any float number greater or equal 0 followed by an optional %
+    """
+    if not re.match(r'^[0-9]+(\.[0-9]+)?(%)?$', field.data):
+        raise ValidationError(_('Invalid PayPal Percent Fee value'))
```
duartegalvao, line 39: No need for these if you use the `DecimalField`

payment_paypal/indico_payment_paypal/plugin.py
```diff
@@ -61,13 +72,27 @@ def get_blueprints(self):
...
+        data['paypal_amount'] = amount
+        paypal_fixed_fee = float(event_settings_form['paypal_fixed_fee'])
+        paypal_percent_fee = float(re.findall(r'^([0-9]+(\.[0-9]+)?)(%)?', event_settings_form['paypal_percent_fee'])[0][0])
+        amount_with_paypal_fees = round (( amount + paypal_fixed_fee ) / ( 1 - ( paypal_percent_fee / 100 )), 2)
+        data['amount'] = amount_with_paypal_fees
+        fees = amount_with_paypal_fees - amount
+        if fees > 0.0:
+            data['paypal_fees'] = str(amount_with_paypal_fees - amount)
+        else:
+            data['paypal_fees'] = ''
```
duartegalvao, line 94: If these are stored as decimals you can just:
```suggestion
        data['paypal_fees'] = str(amount_with_paypal_fees - amount)
```

duartegalvao (changes requested):
thanks for your contribution! there are a few code style issues, I've left my suggestions here

payment_paypal/indico_payment_paypal/controllers.py
```diff
@@ -65,12 +67,14 @@ def _process(self):
                                           payment_status, request.form)
             return
         self._verify_amount()
+        current_plugin.logger.warning("Payment verify ok, now register transaction...")
```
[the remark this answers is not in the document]
rabser: all the suggestions seems fine and the work is almost done,  but applying the DecimalField, I'm getting this error when saving the event option:
```
StatementError: (builtins.TypeError) Object of type Decimal is not JSON serializable [SQL: INSERT INTO events.settings (module, name, value, event_id) VALUES (%(module)s, %(name)s, %(value)s, %(event_id)s) RETURNING events.settings.id] [parameters: [{'event_id': 1, 'value': 'mybusiness@paypal.it', 'module': 'plugin_payment_paypal', 'name': 'business'}, {'event_id': 1, 'value': 'PayPal', 'module': ... (199 characters truncated) ... lugin_payment_paypal', 'name': 'enabled'}, {'event_id': 1, 'value': Decimal('0.00'), 'module': 'plugin_payment_paypal', 'name': 'paypal_percent_fee'}]]
```
The error is clear,  but I'm not so used to coding in python so deep: some hints on how to solve it ? The event.settings in effect admit a jsonb value and the Decimal value seems not serializable. Should I do some casts ? In the indico code I didn't able to find an example.

payment_paypal/indico_payment_paypal/controllers.py
```diff
@@ -5,21 +5,24 @@
...
 from indico.modules.events.payment.models.transactions import TransactionAction
 from indico.modules.events.payment.notifications import notify_amount_inconsistency
 from indico.modules.events.payment.util import register_transaction
 from indico.modules.events.registration.models.registrations import Registration
 from indico.web.flask.util import url_for
 from indico.web.rh import RH
 
+from flask import flash, redirect, request
 from indico_payment_paypal import _
+from werkzeug.exceptions import BadRequest
```
duartegalvao, line 25: this one is still out of order, it should be in the second import group (make sure to run isort before right commiting)

payment_paypal/indico_payment_paypal/controllers.py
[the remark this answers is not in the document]
ThiefMaster: Do we really need to store it as a decimal? Maybe storing it as a number (float) is fine here...

payment_paypal/indico_payment_paypal/controllers.py
```diff
@@ -46,7 +49,7 @@ def _process(self):
         verify_params = list(chain(IPN_VERIFY_EXTRA_PARAMS, request.form.items()))
         result = requests.post(current_plugin.settings.get('url'), data=verify_params).text
         if result != 'VERIFIED':
-            current_plugin.logger.warning('Paypal IPN string %s did not validate (%s)', verify_params, result)
+            current_plugin.logger.debug('Paypal IPN string %s did not validate (%s)', verify_params, result)
```
duartegalvao, line 52: this one should be a warning, since it is indeed a failure...

payment_paypal/indico_payment_paypal/controllers.py
```diff
@@ -84,14 +89,22 @@ def _verify_business(self):
         return False
 
     def _verify_amount(self):
-        expected_amount = self.registration.price
+        paypal_fixed_fee = float(current_plugin.event_settings.get(self.registration.registration_form.event,
+            'paypal_fixed_fee'))
+        paypal_percent_fee = float(current_plugin.event_settings.get(self.registration.registration_form.event,
+            'paypal_percent_fee'))
+        expected_amount = float(self.registration.price)
```
duartegalvao, line 96: avoid using floats for anything payment-related, since these can introduce tiny rounding errors in larger numbers.. please always use the `Decimal` type

payment_paypal/indico_payment_paypal/plugin.py
```diff
@@ -61,13 +75,23 @@ def get_blueprints(self):
...
         plain_title = str_to_ascii(remove_accents(event.title))
         data['item_name'] = f'{plain_name}: registration for {plain_title}'
         data['return_url'] = url_for_plugin('payment_paypal.success', registration.locator.uuid, _external=True)
         data['cancel_url'] = url_for_plugin('payment_paypal.cancel', registration.locator.uuid, _external=True)
         data['notify_url'] = url_for_plugin('payment_paypal.notify', registration.locator.uuid, _external=True)
+        # Add Paypal fees
+        amount = float(data['amount'])
+        data['paypal_amount'] = amount
+        paypal_fixed_fee = float(event_settings_form['paypal_fixed_fee'])
+        paypal_percent_fee = float(event_settings_form['paypal_percent_fee'])
```
duartegalvao, line 90: same here

payment_paypal/indico_payment_paypal/util.py
```diff
@@ -7,11 +7,10 @@
 
 import re
 
-from wtforms import ValidationError
-
 from indico.util.string import is_valid_mail
 
 from indico_payment_paypal import _
+from wtforms import ValidationError
```
duartegalvao, line 13: again here, isort should fix this..

duartegalvao (changes requested):
a few more suggestions.. also feel free to mark as resolved the comments which you have addressed, makes it easier to keep track of things 😄 

EDIT: also, please check the linting errors in the CI, I have just enabled it

payment_paypal/indico_payment_paypal/controllers.py
```diff
@@ -107,13 +120,28 @@ class RHPaypalSuccess(RHPaypalIPN):
     """Confirmation message after successful payment"""
 
     def _process(self):
+        # Force check on transaction
+        sleeps=0
+        registration = Registration.query.filter_by(uuid=self.token).first()
+        while ((not registration.is_paid) and (sleeps < 30)):
+            current_plugin.logger.debug('Waiting for transaction registration %s',sleeps)
+            time.sleep(1)
```
ThiefMaster, line 128: This is not OK. It hogs a uWSGI worker for up to 30s. Sure, the risk of this causing a significant problem is low, but especially no small deployments not negligible.

Check how it was done in [this plugin](https://github.com/indico/indico-plugins-cern/pull/153/files#diff-3bd673336924bee0efbb767ddc620e6119004b1d2b05e965031abe03e1d6bed7) instead.

payment_paypal/indico_payment_paypal/controllers.py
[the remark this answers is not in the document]
ThiefMaster: But indeed, converting from decimal to string for saving and later back from string to decimal is safer.

payment_paypal/indico_payment_paypal/util.py
rabser: I did an isort to all three python files changed before committing. Anyway, it's not the latest release, so let's wait until I can finish them all.

payment_paypal/indico_payment_paypal/controllers.py
rabser: Usually after 2/3 secs, the while passes over, but I understand your idea and approve it. I'll work in such a way.

payment_paypal/indico_payment_paypal/controllers.py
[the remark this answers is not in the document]
rabser: I'm not able to do the cast DecimalField to str (probably getting back from str to Decimal is easier) and I didn't find any example/solution on the web useful. If I cast directly the DecimalField, the Field disappear from the interface, i tried also to subclass the DecimalField, but without luck.
Do you have an example of how to modify the plugin in the EventSettingsForm class to be able to save to the DB (and probably recall from DB) ?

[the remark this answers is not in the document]
duartegalvao: I think the best way would be to override the form's `data` property and cast the decimals there..

something like this:
```python
    @property
    def data(self):
        return {k: str(v) for k, v in super().data.items()}
 ```
(untested code)

when fetching the setting it should be as simple as `Decimal(value)` im pretty sure..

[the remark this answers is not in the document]
Comment (rabser) [reply payment_paypal/indico_payment_paypal/controllers.py]: Indeed it seems to work. Now I have to rewrite the code as suggested by ThiefMaster ... WIP.

File: payment_paypal/indico_payment_paypal/controllers.py [hunk 17]
```diff
@@ -14,13 +14,15 @@
...
+from indico.modules.events.payment.util import register_transaction, TransactionStatus
 from indico.modules.events.registration.models.registrations import Registration
 from indico.web.flask.util import url_for
 from indico.web.rh import RH
 
 from indico_payment_paypal import _
-
+import re
+import time
+from indico.core.db import db
```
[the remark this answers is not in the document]
rabser: done on all .py

payment_paypal/indico_payment_paypal/controllers.py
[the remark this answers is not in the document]
rabser: all done

payment_paypal/indico_payment_paypal/controllers.py
```diff
@@ -65,12 +67,14 @@ def _process(self):
...
             return
         self._verify_amount()
+        current_plugin.logger.warning("Payment verify ok, now register transaction...")
         register_transaction(registration=self.registration,
                              amount=float(request.form['mc_gross']),
                              currency=request.form['mc_currency'],
                              action=paypal_transaction_action_mapping[payment_status],
                              provider='paypal',
                              data=request.form)
+        current_plugin.logger.warning("Payment process finished")
```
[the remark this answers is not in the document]
rabser: done

payment_paypal/indico_payment_paypal/controllers.py
rabser: not needed anymore, done.

payment_paypal/indico_payment_paypal/controllers.py
rabser: should be done

payment_paypal/indico_payment_paypal/controllers.py
rabser: done.

payment_paypal/indico_payment_paypal/controllers.py
rabser: done

payment_paypal/indico_payment_paypal/plugin.py
rabser: done

payment_paypal/indico_payment_paypal/plugin.py
rabser: done

payment_paypal/indico_payment_paypal/util.py
rabser: removed

payment_paypal/indico_payment_paypal/plugin.py
rabser: done

payment_paypal/indico_payment_paypal/controllers.py
rabser: I ran isort before every commit...

payment_paypal/indico_payment_paypal/controllers.py
rabser: done

payment_paypal/indico_payment_paypal/controllers.py
rabser: changed all the logic, following the CERN example.

PR #235, opened 2024-08-09
Add Typescript support in the plugins

pferreir:
Following up on https://github.com/indico/indico/pull/6456


tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+    "extends": "../src/tsconfig.json",
+    "compilerOptions": {
+      "baseUrl": "../src",
+    },
+    "include": [
+        "../src/indico/web/client/js/types",
+        "vc_zoom/indico_vc_zoom/client/**/*",
```
pferreir, line 8: Actually, this is prescient, since we don't have any Typescript in that dir (yet). I'll add it in my coming Zoom Rooms PR. Feel free to remove if it bothers you.

PR #240, opened 2024-09-02
Piwik: Pass auth token via POST

ThiefMaster:


ThiefMaster:
Keeping it as draft for now until I figured out why the segment filters are not working on the CERN instance in case this is something to be fixed here...

PR #259, opened 2025-03-20
Fix room name display in the compact theme

tomasr8:
There are some extra empty parens when the room name is not set.

now:
image

with the fix:
image


ThiefMaster:
Please bump the version by +0.0.1 and add a changelog line in the plugin's readme

PR #292, opened 2025-12-24
VC/Zoom: Add meeting registration sync

moliholy:
This PR enables registration in zoom meetings when they get registered in indico events with an associated zoom meeting.

### DIFF
--- a/vc_zoom/README.md
+++ b/vc_zoom/README.md
@@ -200,6 +200,7 @@ The scopes to select when creating the app are:
 - `meeting:delete:meeting:admin`
 - `meeting:read:list_registrants:admin` (optional, only needed for automatic registration)
 - `meeting:write:registrant:admin` (optional, only needed for automatic registration)
+- `meeting:update:registrant_status:admin` (optional, only needed for automatic registration)
 - `user:read:user:admin`
 
 - `webinar:read:webinar:admin` (optional, only needed when using webinars)
@@ -208,6 +209,7 @@ The scopes to select when creating the app are:
 - `webinar:delete:webinar:admin` (optional, only needed when using webinars)
 - `webinar:read:list_registrants:admin` (optional, required when using automatic registration with webinars)
 - `webinar:write:registrant:admin` (optional, required when using automatic registration with webinars)
+- `webinar:update:registrant_status:admin` (optional, required when using automatic registration with webinars)
 
 In case your app still uses classic scopes (unlikely, unless you created it a long time ago):
 
--- a/vc_zoom/README.md
+++ b/vc_zoom/README.md
@@ -11,6 +11,15 @@
 
 ## Changelog
 
+### 3.3.8
+
+- Add automatic Zoom registration: event managers can opt-in to automatically register/unregister
+  Indico participants in the corresponding Zoom meeting or webinar
+
+### 3.3.7
+
+- Support Zoom's [Language Interpretation](https://support.zoom.com/hc/en/article?id=zm_kb&sysparm_article=KB0064768) feature
+
 ### 3.3.6
 
 - Compare webhook HMAC token in a more secure manner
--- a/vc_zoom/README.md
+++ b/vc_zoom/README.md
@@ -15,8 +15,6 @@
 
 - Add automatic Zoom registration: event managers can opt-in to automatically register/unregister
   Indico participants in the corresponding Zoom meeting or webinar
-- Show the personalized Zoom join URL on the registration summary page so participants
-  without an Indico account can retrieve it from their registration token link
 - Support Zoom's [Language Interpretation](https://support.zoom.com/hc/en/article?id=zm_kb&sysparm_article=KB0064768) feature
 - Compare webhook HMAC token in a more secure manner
 - Generate random meeting passcodes in a more secure manner
--- a/vc_zoom/README.md
+++ b/vc_zoom/README.md
@@ -11,17 +11,11 @@
 
 ## Changelog
 
-### 3.3.8
+### 3.3.6
 
 - Add automatic Zoom registration: event managers can opt-in to automatically register/unregister
   Indico participants in the corresponding Zoom meeting or webinar
-
-### 3.3.7
-
 - Support Zoom's [Language Interpretation](https://support.zoom.com/hc/en/article?id=zm_kb&sysparm_article=KB0064768) feature
-
-### 3.3.6
-
 - Compare webhook HMAC token in a more secure manner
 - Generate random meeting passcodes in a more secure manner
 - Support Zoom's [Language Interpretation](https://support.zoom.com/hc/en/article?id=zm_kb&sysparm_article=KB0064768) feature
--- a/vc_zoom/README.md
+++ b/vc_zoom/README.md
@@ -186,12 +186,16 @@ The scopes to select when creating the app are:
 - `meeting:write:meeting:admin`
 - `meeting:update:meeting:admin`
 - `meeting:delete:meeting:admin`
+- `meeting:read:list_registrants:admin` (required for automatic registration feature)
+- `meeting:write:registrant:admin` (required for automatic registration feature)
 - `user:read:user:admin`
 
 - `webinar:read:webinar:admin` (optional, only needed when using webinars)
 - `webinar:write:webinar:admin` (optional, only needed when using webinars)
 - `webinar:update:webinar:admin` (optional, only needed when using webinars)
 - `webinar:delete:webinar:admin` (optional, only needed when using webinars)
+- `webinar:read:list_registrants:admin` (optional, required when using automatic registration with webinars)
+- `webinar:write:registrant:admin` (optional, required when using automatic registration with webinars)
 
 In case your app still uses classic scopes (unlikely, unless you created it a long time ago):
 
--- a/vc_zoom/README.md
+++ b/vc_zoom/README.md
@@ -15,6 +15,8 @@
 
 - Add automatic Zoom registration: event managers can opt-in to automatically register/unregister
   Indico participants in the corresponding Zoom meeting or webinar
+- Show the personalized Zoom join URL on the registration summary page so participants
+  without an Indico account can retrieve it from their registration token link
 - Support Zoom's [Language Interpretation](https://support.zoom.com/hc/en/article?id=zm_kb&sysparm_article=KB0064768) feature
 - Compare webhook HMAC token in a more secure manner
 - Generate random meeting passcodes in a more secure manner
--- a/vc_zoom/README.md
+++ b/vc_zoom/README.md
@@ -173,6 +173,9 @@ These are the most relevant configuration options:
  * **Notification email addresses** - Additional e-mails which will receive notifications
  * **E-mail domains** - List of e-mail domains which can be used for the Zoom API (e.g. `cern.ch`)
  * **Webhook Secret Token** (optional) - the token which Zoom requests will authenticate with (get it from Zoom Marketplace)
+ * **Allow automatic registration** - Enable this to allow event managers to opt-in to automatic Zoom
+   registration on individual meetings/webinars. Requires the registration-related scopes listed below.
+   When enabled, a per-room "Automatic registration" toggle becomes available in the meeting settings.
 
 
 ### Zoom Server-to-Server OAuth
@@ -186,8 +189,8 @@ The scopes to select when creating the app are:
 - `meeting:write:meeting:admin`
 - `meeting:update:meeting:admin`
 - `meeting:delete:meeting:admin`
-- `meeting:read:list_registrants:admin` (required for automatic registration feature)
-- `meeting:write:registrant:admin` (required for automatic registration feature)
+- `meeting:read:list_registrants:admin` (optional, only needed for automatic registration)
+- `meeting:write:registrant:admin` (optional, only needed for automatic registration)
 - `user:read:user:admin`
 
 - `webinar:read:webinar:admin` (optional, only needed when using webinars)
--- a/vc_zoom/README.md
+++ b/vc_zoom/README.md
@@ -13,13 +13,12 @@
 
 ### 3.3.6
 
-- Add automatic Zoom registration: event managers can opt-in to automatically register/unregister
-  Indico participants in the corresponding Zoom meeting or webinar
-- Support Zoom's [Language Interpretation](https://support.zoom.com/hc/en/article?id=zm_kb&sysparm_article=KB0064768) feature
 - Compare webhook HMAC token in a more secure manner
 - Generate random meeting passcodes in a more secure manner
 - Support Zoom's [Language Interpretation](https://support.zoom.com/hc/en/article?id=zm_kb&sysparm_article=KB0064768) feature
 - Never use a PMI meeting ID when creating a new meeting
+- Add automatic Zoom registration: event managers can opt-in to automatically register/unregister
+  Indico participants in the corresponding Zoom meeting or webinar
 
 ### 3.3.5
 
--- a/vc_zoom/README.md
+++ b/vc_zoom/README.md
@@ -194,6 +194,7 @@ The scopes to select when creating the app are:
 - `meeting:delete:meeting:admin`
 - `meeting:read:list_registrants:admin` (optional, only needed for automatic registration)
 - `meeting:write:registrant:admin` (optional, only needed for automatic registration)
+- `meeting:write:batch_registrants:admin` (optional, only needed for automatic registration)
 - `meeting:update:registrant_status:admin` (optional, only needed for automatic registration)
 - `user:read:user:admin`
 
@@ -203,6 +204,7 @@ The scopes to select when creating the app are:
 - `webinar:delete:webinar:admin` (optional, only needed when using webinars)
 - `webinar:read:list_registrants:admin` (optional, required when using automatic registration with webinars)
 - `webinar:write:registrant:admin` (optional, required when using automatic registration with webinars)
+- `webinar:write:batch_registrants:admin` (optional, required when using automatic registration with webinars)
 - `webinar:update:registrant_status:admin` (optional, required when using automatic registration with webinars)
 
 In case your app still uses classic scopes (unlikely, unless you created it a long time ago):
--- a/vc_zoom/indico_vc_zoom/api/client.py
+++ b/vc_zoom/indico_vc_zoom/api/client.py
@@ -277,7 +277,8 @@ class ZoomIndicoClient:
         return _handle_response(self.client.meeting.add_registrant(meeting_id, **data), 201)
 
     def update_meeting_registrants_status(self, meeting_id, data):
-        return _handle_response(self.client.meeting.update_registrants_status(meeting_id, **data), 204, expects_json=False)
+        return _handle_response(self.client.meeting.update_registrants_status(meeting_id, **data), 204,
+                                expects_json=False)
 
     def create_webinar(self, user_id, **kwargs):
         return _handle_response(self.client.webinar.create(user_id, **kwargs), 201)
--- a/vc_zoom/indico_vc_zoom/api/client.py
+++ b/vc_zoom/indico_vc_zoom/api/client.py
@@ -110,6 +110,11 @@ class MeetingComponent(BaseComponent):
             f'{self.base_uri}/meetings/{meeting_id}', json=kwargs
         )
 
+    def list_registrants(self, meeting_id, **kwargs):
+        return self.session.get(
+            f'{self.base_uri}/meetings/{meeting_id}/registrants', params=kwargs
+        )
+
     def get_registrant(self, meeting_id, registrant_id):
         return self.session.get(
             f'{self.base_uri}/meetings/{meeting_id}/registrants/{registrant_id}'
@@ -158,6 +163,11 @@ class WebinarComponent(BaseComponent):
     def list_registrants(self, webinar_id, **kwargs):
         return self.session.get(
             f'{self.base_uri}/webinars/{webinar_id}/registrants', params=kwargs
+        )
+
+    def get_registrant(self, webinar_id, registrant_id):
+        return self.session.get(
+            f'{self.base_uri}/webinars/{webinar_id}/registrants/{registrant_id}'
         )
 
     def add_registrant(self, webinar_id, **kwargs):
@@ -260,8 +270,8 @@ class ZoomIndicoClient:
     def delete_meeting(self, meeting_id):
         return _handle_response(self.client.meeting.delete(meeting_id), 204, expects_json=False)
 
-    def list_meeting_registrants(self, meeting_id, registrant_id):
-        return _handle_response(self.client.meeting.get_registrant(meeting_id, registrant_id))
+    def list_meeting_registrants(self, meeting_id, **kwargs):
+        return _handle_response(self.client.meeting.list_registrants(meeting_id, **kwargs))
 
     def add_meeting_registrant(self, meeting_id, data):
         return _handle_response(self.client.meeting.add_registrant(meeting_id, **data), 201)
@@ -281,7 +291,10 @@ class ZoomIndicoClient:
     def delete_webinar(self, webinar_id):
         return _handle_response(self.client.webinar.delete(webinar_id), 204, expects_json=False)
 
-    def list_webinar_registrants(self, webinar_id, registrant_id):
+    def list_webinar_registrants(self, webinar_id, **kwargs):
+        return _handle_response(self.client.webinar.list_registrants(webinar_id, **kwargs))
+
+    def get_webinar_registrant(self, webinar_id, registrant_id):
         return _handle_response(self.client.webinar.get_registrant(webinar_id, registrant_id))
 
     def add_webinar_registrant(self, webinar_id, data):
--- a/vc_zoom/indico_vc_zoom/api/client.py
+++ b/vc_zoom/indico_vc_zoom/api/client.py
@@ -110,6 +110,21 @@ class MeetingComponent(BaseComponent):
             f'{self.base_uri}/meetings/{meeting_id}', json=kwargs
         )
 
+    def get_registrant(self, meeting_id, registrant_id):
+        return self.session.get(
+            f'{self.base_uri}/meetings/{meeting_id}/registrants/{registrant_id}'
+        )
+
+    def add_registrant(self, meeting_id, **kwargs):
+        return self.session.post(
+            f'{self.base_uri}/meetings/{meeting_id}/registrants', json=kwargs
+        )
+
+    def update_registrants_status(self, meeting_id, **kwargs):
+        return self.session.put(
+            f'{self.base_uri}/meetings/{meeting_id}/registrants/status', json=kwargs
+        )
+
 
 class WebinarComponent(BaseComponent):
     def list(self, user_id, **kwargs):
@@ -138,6 +153,21 @@ class WebinarComponent(BaseComponent):
     def delete(self, meeting_id, **kwargs):
         return self.session.delete(
             f'{self.base_uri}/webinars/{meeting_id}', json=kwargs
+        )
+
+    def list_registrants(self, webinar_id, **kwargs):
+        return self.session.get(
+            f'{self.base_uri}/webinars/{webinar_id}/registrants', params=kwargs
+        )
+
+    def add_registrant(self, webinar_id, **kwargs):
+        return self.session.post(
+            f'{self.base_uri}/webinars/{webinar_id}/registrants', json=kwargs
+        )
+
+    def update_registrants_status(self, webinar_id, **kwargs):
+        return self.session.put(
+            f'{self.base_uri}/webinars/{webinar_id}/registrants/status', json=kwargs
         )
 
 
@@ -230,6 +260,15 @@ class ZoomIndicoClient:
     def delete_meeting(self, meeting_id):
         return _handle_response(self.client.meeting.delete(meeting_id), 204, expects_json=False)
 
+    def list_meeting_registrants(self, meeting_id, registrant_id):
+        return _handle_response(self.client.meeting.get_registrant(meeting_id, registrant_id))
+
+    def add_meeting_registrant(self, meeting_id, data):
+        return _handle_response(self.client.meeting.add_registrant(meeting_id, **data), 201)
+
+    def update_meeting_registrants_status(self, meeting_id, data):
+        return _handle_response(self.client.meeting.update_registrants_status(meeting_id, **data), 204, expects_json=False)
+
     def create_webinar(self, user_id, **kwargs):
         return _handle_response(self.client.webinar.create(user_id, **kwargs), 201)
 
@@ -241,6 +280,16 @@ class ZoomIndicoClient:
 
     def delete_webinar(self, webinar_id):
         return _handle_response(self.client.webinar.delete(webinar_id), 204, expects_json=False)
+
+    def list_webinar_registrants(self, webinar_id, registrant_id):
+        return _handle_response(self.client.webinar.get_registrant(webinar_id, registrant_id))
+
+    def add_webinar_registrant(self, webinar_id, data):
+        return _handle_response(self.client.webinar.add_registrant(webinar_id, **data), 201)
+
+    def update_webinar_registrants_status(self, webinar_id, data):
+        return _handle_response(self.client.webinar.update_registrants_status(webinar_id, **data), 204,
+                                expects_json=False)
 
     def get_user(self, user_id, silent=False):
         resp = self.client.user.get(user_id)
--- a/vc_zoom/indico_vc_zoom/api/client.py
+++ b/vc_zoom/indico_vc_zoom/api/client.py
@@ -343,22 +343,16 @@ def get_zoom_scopes(config):
         return set()
 
     try:
-        access_token, *_rest = get_zoom_token(config, for_config_check=True)
-    except Exception:
+        token_data = get_zoom_token(config, force=True, full=True)
+    except requests.RequestException:
         return set()
-    if not access_token:
-        return set()
-
-    # get_zoom_token populates the token_cache as a side effect;
-    # read it back to access the scope field, which is not in the return tuple.
-    token_data = token_cache.get(_get_token_cache_key(config))
     if not token_data:
         return set()
 
     return set(token_data.get('scope', '').split())
 
 
-def get_zoom_token(config, *, force=False, for_config_check=False):
+def get_zoom_token(config, *, force=False, for_config_check=False, full=False):
     from indico_vc_zoom.plugin import ZoomPlugin
 
     account_id = config['account_id']
@@ -373,6 +367,8 @@ def get_zoom_token(config, *, force=False, for_config_check=False):
     if not force and (token_data := token_cache.get(cache_key)):
         expires_in = int(token_data['expires_at'] - time.time())
         ZoomPlugin.logger.debug('Using token from cache (%s, %ds remaining)', cache_key, expires_in)
+        if full:
+            return token_data
         if for_config_check:
             return True, token_data['api_url'], 'cached'
         return token_data['access_token'], token_data['api_url'], token_data['expires_at']
@@ -384,6 +380,8 @@ def get_zoom_token(config, *, force=False, for_config_check=False):
         )
         resp.raise_for_status()
     except HTTPError as exc:
+        if full:
+            return None
         if for_config_check:
             return False, None, exc.response.text
         ZoomPlugin.logger.error('Could not get zoom token: %s', exc.response.text if exc.response else exc)
@@ -395,6 +393,8 @@ def get_zoom_token(config, *, force=False, for_config_check=False):
     expires_at = int(time.time() + token_data['expires_in'])
     token_data.setdefault('expires_at', expires_at)  # zoom doesn't include this. wtf.
     token_cache.set(cache_key, token_data, token_data['expires_in'])
+    if full:
+        return token_data
     if for_config_check:
         return True, token_data['api_url'], 'fresh'
     return token_data['access_token'], token_data['api_url'], token_data['expires_at']
--- a/vc_zoom/indico_vc_zoom/api/client.py
+++ b/vc_zoom/indico_vc_zoom/api/client.py
@@ -20,6 +20,11 @@ from indico.util.string import crc32
 token_cache = make_scoped_cache('zoom-api-token')
 
 
+def _get_token_cache_key(config):
+    cache_key_source = '-'.join((config['account_id'], config['client_id'], config['client_secret']))
+    return f'token-{crc32(cache_key_source)}'
+
+
 def format_iso_dt(d):
     """Convert a datetime objects to a UTC-based string.
 
@@ -312,6 +317,31 @@ class ZoomIndicoClient:
         return _handle_response(resp)
 
 
+def get_zoom_scopes(config):
+    """Get the set of OAuth scopes for the given Zoom credentials."""
+    account_id = config['account_id']
+    client_id = config['client_id']
+    client_secret = config['client_secret']
+
+    if not (account_id and client_id and client_secret):
+        return set()
+
+    try:
+        access_token, *_rest = get_zoom_token(config, for_config_check=True)
+    except Exception:
+        return set()
+    if not access_token:
+        return set()
+
+    # get_zoom_token populates the token_cache as a side effect;
+    # read it back to access the scope field, which is not in the return tuple.
+    token_data = token_cache.get(_get_token_cache_key(config))
+    if not token_data:
+        return set()
+
+    return set(token_data.get('scope', '').split())
+
+
 def get_zoom_token(config, *, force=False, for_config_check=False):
     from indico_vc_zoom.plugin import ZoomPlugin
 
@@ -323,8 +353,7 @@ def get_zoom_token(config, *, force=False, for_config_check=False):
         raise Exception('Zoom authentication not configured')
 
     ZoomPlugin.logger.debug(f'Using Server-to-Server-OAuth ({force=})')
-    hash_key = '-'.join((account_id, client_id, client_secret))
-    cache_key = f'token-{crc32(hash_key)}'
+    cache_key = _get_token_cache_key(config)
     if not force and (token_data := token_cache.get(cache_key)):
         expires_in = int(token_data['expires_at'] - time.time())
         ZoomPlugin.logger.debug('Using token from cache (%s, %ds remaining)', cache_key, expires_in)
--- a/vc_zoom/indico_vc_zoom/api/client.py
+++ b/vc_zoom/indico_vc_zoom/api/client.py
@@ -130,6 +130,11 @@ class MeetingComponent(BaseComponent):
             f'{self.base_uri}/meetings/{meeting_id}/registrants', json=kwargs
         )
 
+    def batch_registrants(self, meeting_id, **kwargs):
+        return self.session.post(
+            f'{self.base_uri}/meetings/{meeting_id}/batch_registrants', json=kwargs
+        )
+
     def update_registrants_status(self, meeting_id, **kwargs):
         return self.session.put(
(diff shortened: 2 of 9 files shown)


moliholy:
@ThiefMaster friendly ping 🙏

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -91,6 +91,11 @@ class PluginSettingsForm(VCPluginSettingsFormBase):
                                                  description=_('Allow enabling language interpretation for meetings '
                                                                'and webinars.'))
 
+    auto_register = BooleanField(_('Automatic registration'),
+                                 widget=SwitchWidget(),
+                                 description=_('Automatically register Indico registrants in Zoom meetings/webinars '
+                                               'once their Indico registration is complete.'))
```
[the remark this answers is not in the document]
moliholy: Done in [4bb83393](https://github.com/indico/indico-plugins/pull/292/commits/4bb83393).

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -621,3 +630,76 @@ def _event_metadata_postprocess(self, sender, event, data, user=None, skip_acces
...
+        should_be_registered = not remove and is_active
+
+        client = ZoomIndicoClient()
+        for vc_room in zoom_rooms:
+            zoom_id = vc_room.data['zoom_id']
+            is_webinar = vc_room.data.get('meeting_type') == 'webinar'
+
+            if should_be_registered:
+                data = {
+                    'email': registration.email,
```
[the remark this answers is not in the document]
moliholy: Done in [d58b4ff2](https://github.com/indico/indico-plugins/pull/292/commits/d58b4ff2).

vc_zoom/indico_vc_zoom/api/client.py
```diff
@@ -242,13 +310,54 @@ def update_webinar(self, webinar_id, data):
...
+    client_id = config['client_id']
+    client_secret = config['client_secret']
+
+    if not (account_id and client_id and client_secret):
+        return set()
+
+    try:
+        access_token, *_rest = get_zoom_token(config, for_config_check=True)
+    except Exception:
+        return set()
```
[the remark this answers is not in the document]
moliholy: I guess only a network error. Fixed in [e1117a2](https://github.com/indico/indico-plugins/pull/292/commits/e1117a26f7468ef9a5c2b2a8bce9a0aef71a4a7d).

vc_zoom/indico_vc_zoom/api/client.py
```diff
@@ -242,13 +310,54 @@ def update_webinar(self, webinar_id, data):
...
+    if not access_token:
+        return set()
+
+    # get_zoom_token populates the token_cache as a side effect;
+    # read it back to access the scope field, which is not in the return tuple.
```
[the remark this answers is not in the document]
moliholy: Was thinking about that, actually. Done in [e1117a2](https://github.com/indico/indico-plugins/pull/292/commits/e1117a26f7468ef9a5c2b2a8bce9a0aef71a4a7d).

vc_zoom/README.md
```diff
@@ -188,12 +188,16 @@ The scopes to select when creating the app are:
 - `meeting:write:meeting:admin`
 - `meeting:update:meeting:admin`
 - `meeting:delete:meeting:admin`
+- `meeting:read:list_registrants:admin` (required for automatic registration feature)
+- `meeting:write:registrant:admin` (required for automatic registration feature)
```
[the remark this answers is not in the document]
moliholy: Done in [e1117a2](https://github.com/indico/indico-plugins/pull/292/commits/e1117a26f7468ef9a5c2b2a8bce9a0aef71a4a7d)

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -129,18 +170,37 @@ def validate_authenticators(self, field):
...
+        if not all(config.values()):
             flash(_('Zoom credentials not set; the plugin will not work correctly'), 'error')
             return
 
-        ok, url, msg = get_zoom_token({'account_id': account_id, 'client_id': client_id,
-                                       'client_secret': client_secret},
-                                      for_config_check=True)
-        if ok:
+        access_token, url, msg = get_zoom_token(config, for_config_check=True)
+        if access_token:
```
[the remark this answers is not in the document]
moliholy: Done in [e1117a2](https://github.com/indico/indico-plugins/pull/292/commits/e1117a26f7468ef9a5c2b2a8bce9a0aef71a4a7d).

vc_zoom/indico_vc_zoom/api/client.py
```diff
@@ -242,13 +310,54 @@ def update_webinar(self, webinar_id, data):
...
+    """Get the set of OAuth scopes for the given Zoom credentials."""
+    account_id = config['account_id']
+    client_id = config['client_id']
+    client_secret = config['client_secret']
+
+    if not (account_id and client_id and client_secret):
+        return set()
+
+    try:
+        access_token, *_rest = get_zoom_token(config, for_config_check=True)
```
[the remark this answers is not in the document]
moliholy: Yep, indeed. Done in [e1117a2](https://github.com/indico/indico-plugins/pull/292/commits/e1117a26f7468ef9a5c2b2a8bce9a0aef71a4a7d).

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -91,6 +122,16 @@ class PluginSettingsForm(VCPluginSettingsFormBase):
                                                  description=_('Allow enabling language interpretation for meetings '
                                                                'and webinars.'))
 
+    auto_register = BooleanField(_('Automatic registration'),
+                                 widget=SwitchWidget(),
+                                 description=_('Automatically register Indico registrants in Zoom meetings/webinars '
+                                               'once their Indico registration is complete. Requires the Zoom API '
+                                               'scopes {}. If webinars are enabled, add {} as well. Consider setting '
+                                               '"Passcode visibility" to "No one" so participants '
+                                               'only receive their personalized join link.')
```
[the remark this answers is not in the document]
moliholy: OK, I see. This a a bigger one. Done in [a52803d](https://github.com/indico/indico-plugins/pull/292/commits/a52803d3583bad413ac45ea7f3c00d224b8c940e). I set it as enabled by default (if visible), and I'd recommend setting by default the passcode visibility to "No one" by default in this case too (not done, just a suggestion).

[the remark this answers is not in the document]
ThiefMaster: my preference would be to default to off, also because many instances will have mainly meetings where "registrants" are used merely to track participants (but no open registration), and there the feature would not be ideal

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -129,18 +168,37 @@ def validate_authenticators(self, field):
...
+        }
+
+    def validate_allow_auto_register(self, field):
+        if not field.data:
+            return
+        config = self._get_zoom_config()
+        if not all(config.values()):
+            return
+        scopes = get_zoom_scopes(config)
+        if not scopes:
```
ThiefMaster:
```suggestion
        if not (scopes := get_zoom_scopes(config)):
```

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -129,18 +168,37 @@ def validate_authenticators(self, field):
...
+            return
+        missing_scopes = _get_missing_auto_registration_scopes(scopes, allow_webinars=self.allow_webinars.data)
+        if missing_scopes:
+            raise ValidationError(
```
ThiefMaster, line 204:
```suggestion
        if missing_scopes := _get_missing_auto_registration_scopes(scopes, allow_webinars=self.allow_webinars.data)
```

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -177,12 +236,21 @@ class ZoomPlugin(VCPluginMixin, IndicoPlugin):
         'send_host_url': False,
         'phone_link': '',
     }
+    # Limited to 30 by Zoom API.
+    # See https://developers.zoom.us/docs/api/meetings/#tag/invitation--registration/post/meetings/{meetingId}/registrants
+    _BATCH_REGISTRANTS_MAX = 30
```
ThiefMaster: I would set this on the module level together with the other (scope-related) constants and not inside the plugin class. Or alternatively on the zoom client class where it would IMHO fit even better.

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -665,3 +737,234 @@ def _event_metadata_postprocess(self, sender, event, data, user=None, skip_acces
...
+                .first())
+
+    def get_personalized_join_url(self, vc_room, event_vc_room, user):
+        if user is None or vc_room.type != self.service_name or not self.settings.get('allow_auto_register'):
+            return None
+        if not vc_room.data.get('auto_register'):
+            return None
+
+        cache_key = None
+        if has_request_context():
```
ThiefMaster: Is this ever called outside a request context? (maybe when getting the zoom link for emails or ical?)

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -665,3 +737,234 @@ def _event_metadata_postprocess(self, sender, event, data, user=None, skip_acces
...
+        pending[registration.id] = (registration, remove)
+
+    def _flush_pending_registrations(self, sender, **kwargs):
+        pending = g.pop('zoom_pending_registrations', None)
+        if not pending:
+            return
+
+        room_ops = self._collect_room_ops(pending)
+        if not room_ops:
+            return
```
ThiefMaster:
```suggestion
        if not (pending := g.pop('zoom_pending_registrations', None)):
            return

        if not (room_ops := self._collect_room_ops(pending)):
            return
```

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -665,3 +737,234 @@ def _event_metadata_postprocess(self, sender, event, data, user=None, skip_acces
...
+        for ops in room_ops.values():
+            ops['add'] = list(ops['add'].values())
+        return room_ops
+
+    def _has_other_active_room_registration(self, vc_room, email, exclude_ids):
+        from indico.modules.events.registration.models.forms import RegistrationForm
+        from indico.modules.events.registration.models.registrations import Registration
+
+        event_ids = {assoc.event_id for assoc in vc_room.events}
+        if not event_ids:
```
ThiefMaster:
```suggestion
        if not (event_ids := {assoc.event_id for assoc in vc_room.events}):
```

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -665,3 +737,234 @@ def _event_metadata_postprocess(self, sender, event, data, user=None, skip_acces
...
+            return
+        for regform in event.registration_forms:
+            for registration in regform.active_registrations:
+                if registration.state == RegistrationState.complete:
+                    self._queue_registration_sync(registration, remove=False)
+
+    def _get_registrant_email(self, registration):
+        if registration.user:
+            enterprise_email = find_enterprise_email(registration.user)
+            if enterprise_email:
```
ThiefMaster:
```suggestion
            if enterprise_email := find_enterprise_email(registration.user):
```

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -665,3 +737,234 @@ def _event_metadata_postprocess(self, sender, event, data, user=None, skip_acces
...
+                return enterprise_email
```
ThiefMaster:
```suggestion
                return enterprise_email.lower()
```

and remove the `email = email.lower()` in all the other places where it's no longer needed.

because on Registration, User, etc. we already enforce lowercase emails in the DB via a check constraint. so the only place where we might end up with mixed-case emails is data coming from zoom (technically it should also not happen there in case of registrations because these come from indico, but for zoom user emails there's no such control)

vc_zoom/README.md
ThiefMaster: Add a changelog line for this feature

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -564,17 +634,19 @@ def get_blueprints(self):
...
     def get_vc_room_form_defaults(self, event):
         defaults = super().get_vc_room_form_defaults(event)
+        auto_register_enabled = self.settings.get('allow_auto_register')
         defaults.update({
             'meeting_type': 'regular' if self.settings.get('allow_webinars') else None,
             'mute_audio': self.settings.get('mute_audio'),
             'mute_host_video': self.settings.get('mute_host_video'),
             'mute_participant_video': self.settings.get('mute_participant_video'),
             'waiting_room': self.settings.get('waiting_room'),
+            'auto_register': auto_register_enabled,
```
ThiefMaster: I still prefer this to be opt-in here

```suggestion
            'auto_register': False,
```

vc_zoom/indico_vc_zoom/plugin.py
ThiefMaster, line 753: Should this also take personalized links into account? This is used when someone generates an ical file for an event.

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -42,11 +43,41 @@
                                  process_alternative_hosts, update_zoom_meeting)
 
 
+AUTO_REGISTRATION_MEETING_SCOPES = ('meeting:read:list_registrants:admin', 'meeting:write:registrant:admin')
```
ThiefMaster: Some scopes are missing here. When withdrawing a registration I got this error:

```
2026-04-13 11:23:44,573  ERROR    a10439468c6a43df  6       indico.plugin.vc_zoom     Error in API call to https://cern.zoom.us/v2/meetings/xxx/registrants/status: b'{"code":4711,"message":"Invalid access token, does not contain scopes:[meeting:update:registrant_status, meeting:update:registrant_status:admin]."}'
```

So I think `meeting:update:registrant_status:admin` needs to be added to the list of required scopes.

ThiefMaster: And probably `webinar:update:registrant_status:admin` as well for webinars.

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -42,11 +43,41 @@
...
+
+
+def _has_required_zoom_scopes(granted_scopes, scope_options):
+    return any(required_scopes <= granted_scopes for required_scopes in scope_options)
+
+
+def _get_missing_auto_registration_scopes(granted_scopes, *, allow_webinars):
+    missing_scopes = []
+    if not _has_required_zoom_scopes(granted_scopes, AUTO_REGISTRATION_SCOPE_OPTIONS['meeting']):
+        missing_scopes.extend(AUTO_REGISTRATION_MEETING_SCOPES)
```
ThiefMaster: Shouldn't this only add the missing scopes to the list, not all required scopes? When I tested adding the status update scopes I mentioned in my last comment, I noticed that I got this error which also includes scopes that are already present:

Image

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -91,6 +122,14 @@ class PluginSettingsForm(VCPluginSettingsFormBase):
                                                  description=_('Allow enabling language interpretation for meetings '
                                                                'and webinars.'))
 
+    allow_auto_register = BooleanField(_('Allow automatic registration'),
+                                      widget=SwitchWidget(),
+                                      description=_('Allow event managers to enable automatic Zoom registration on '
+                                                    'individual meetings/webinars. Requires the Zoom API scopes {}. '
+                                                    'If webinars are enabled, add {} as well.')
+                                      .format(_format_zoom_scopes(AUTO_REGISTRATION_MEETING_SCOPES),
+                                              _format_zoom_scopes(AUTO_REGISTRATION_WEBINAR_SCOPES)))
```
ThiefMaster: Maybe this is overkill to list here. There's the README and if you try to enable this w/o the required scope, you already get an error. A [link to the relevant section of the plugin's README](https://github.com/indico/indico-plugins/tree/master/vc_zoom#zoom-server-to-server-oauth) could be a nice alternative?

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -665,3 +737,234 @@ def _event_metadata_postprocess(self, sender, event, data, user=None, skip_acces
...
+        event_ids = {assoc.event_id for assoc in vc_room.events}
+        if not event_ids:
+            return False
+
+        query = (Registration.query
+                 .join(Registration.registration_form)
+                 .filter(RegistrationForm.event_id.in_(event_ids),
+                         Registration.state == RegistrationState.complete,
+                         ~Registration.is_deleted,
+                         ~RegistrationForm.is_deleted))
```
ThiefMaster: I think you need to join `Registration.event` (or `RegistrationForm.event`, shouldn't matter which one) here as well and include `~Event.is_deleted` in the filters.

vc_zoom/indico_vc_zoom/forms.py
```diff
@@ -106,6 +106,13 @@ class VCRoomForm(VCRoomFormBase):
                                 widget=SwitchWidget(),
                                 description=_('Participants may be kept in a waiting room by the host'))
 
+    auto_register = BooleanField(_('Automatic registration'),
```
ThiefMaster, line 109: I think this option should be disabled (and forced to be enabled) if the Zoom meeting is attached to another event where it's enabled.

Alternatively, we could simply disable the option altogether when adding a zoom meeting to an event that's already attached to other events, ie only allow toggling this when creating.

I would also consider to not allow attaching zoom meeting to an event if registration is enabled. Because we get so many edge cases from this that I think it's just not worth supporting... I can't come up with a good reason why someone would want to re-use a zoom meeting with registration in multiple events tbh.

vc_zoom/indico_vc_zoom/plugin.py
ThiefMaster: I tested on our tenant and adding `meeting:update:registrant_status:admin` fixed the error for meetings.

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -665,3 +737,234 @@ def _event_metadata_postprocess(self, sender, event, data, user=None, skip_acces
...
+
+    def _add_single_registrant(self, client, zoom_id, reg_data, is_webinar):
+        data = {**reg_data, 'auto_approve': True}
+        try:
+            if is_webinar:
+                client.add_webinar_registrant(zoom_id, data)
+            else:
+                client.add_meeting_registrant(zoom_id, data)
+        except HTTPError:
+            self.logger.warning('Could not add registrant %s to Zoom %s %s',
```
ThiefMaster: Logging errors is great, bit I'd also log successes (`info` level). Otherwise it'll be a pain to debug if someone claims their personal Zoom link doesn't work - but with logging it would be trivial to match the removal (or addition) to the action on Indico that triggered it.

ThiefMaster: I would interpolate the webinar/meeting string using an f-string instead of making it a log message argument. That way tools like Sentry will see it as distinct log messages which IMHO makes sense here.


```py
zoom_type = 'webinar' if is_webinar else 'meeting'
self.logger.warning(f'Could not add registrant %s to Zoom {zoom_type} %s', ...)
```

Note that you probably need to silence a ruff warning when doing this, since *usually* it's considered a bad practice. I see it more like that dynamic parts should NOT he interpolated (because you don't want tons of distinct log messages on Sentry), but parts that are quite static ("webinar/meeting") actually make sense to interpolate.

vc_zoom/indico_vc_zoom/plugin.py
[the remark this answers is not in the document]
moliholy: Done in [ab80d76f](https://github.com/indico/indico-plugins/pull/292/commits/ab80d76f15097cc6b01936e22272eca1be00b195).

vc_zoom/indico_vc_zoom/plugin.py
moliholy: Done in [ab80d76f](https://github.com/indico/indico-plugins/pull/292/commits/ab80d76f15097cc6b01936e22272eca1be00b195).

vc_zoom/indico_vc_zoom/plugin.py
moliholy: Done in [ab80d76f](https://github.com/indico/indico-plugins/pull/292/commits/ab80d76f15097cc6b01936e22272eca1be00b195).

vc_zoom/indico_vc_zoom/plugin.py
moliholy: Done in [ab80d76f](https://github.com/indico/indico-plugins/pull/292/commits/ab80d76f15097cc6b01936e22272eca1be00b195).

vc_zoom/indico_vc_zoom/plugin.py
moliholy: Done in [ab80d76f](https://github.com/indico/indico-plugins/pull/292/commits/ab80d76f15097cc6b01936e22272eca1be00b195).

vc_zoom/indico_vc_zoom/plugin.py
moliholy: Done in [ab80d76f](https://github.com/indico/indico-plugins/pull/292/commits/ab80d76f15097cc6b01936e22272eca1be00b195).

vc_zoom/indico_vc_zoom/plugin.py
moliholy: Done in [ab80d76f](https://github.com/indico/indico-plugins/pull/292/commits/ab80d76f15097cc6b01936e22272eca1be00b195).

vc_zoom/indico_vc_zoom/plugin.py
moliholy: Done in [ab80d76f](https://github.com/indico/indico-plugins/pull/292/commits/ab80d76f15097cc6b01936e22272eca1be00b195).

vc_zoom/README.md
moliholy: Done in [ab80d76f](https://github.com/indico/indico-plugins/pull/292/commits/ab80d76f15097cc6b01936e22272eca1be00b195).

vc_zoom/indico_vc_zoom/plugin.py
moliholy: Done in [ab80d76f](https://github.com/indico/indico-plugins/pull/292/commits/ab80d76f15097cc6b01936e22272eca1be00b195).

vc_zoom/indico_vc_zoom/plugin.py
moliholy: Done in ~~[ab80d76f](https://github.com/indico/indico-plugins/pull/292/commits/ab80d76f15097cc6b01936e22272eca1be00b195)~~ [97414af](https://github.com/indico/indico-plugins/pull/292/commits/97414afb30114546a1dd2252ef647baecaf89459).

vc_zoom/indico_vc_zoom/plugin.py
moliholy: Done in [ab80d76f](https://github.com/indico/indico-plugins/pull/292/commits/ab80d76f15097cc6b01936e22272eca1be00b195).

vc_zoom/indico_vc_zoom/plugin.py
moliholy: Yes, it's now also called from the iCal export path.

vc_zoom/indico_vc_zoom/plugin.py
moliholy: Nice catch. Done in [bc9171e](https://github.com/indico/indico-plugins/pull/292/commits/bc9171e235346a31e9d950c4622cbf3006674525).

vc_zoom/indico_vc_zoom/plugin.py
moliholy: Heh, my bad: I already had that scope by mistake in the zoom app I was testing against. Fixed in [bacc034](https://github.com/indico/indico-plugins/pull/292/commits/bacc0340689c66edfcb0be5800d7be8c1335d858).

vc_zoom/indico_vc_zoom/plugin.py
moliholy: You're right, that makes much more sense. Done in [42af8f2](https://github.com/indico/indico-plugins/pull/292/commits/42af8f21f93ca716ad07273e56d3470a63b7b192).

vc_zoom/indico_vc_zoom/plugin.py
moliholy: Yep, makes sense. Done in [ab1af66](https://github.com/indico/indico-plugins/pull/292/commits/ab1af669b056f91f120aa17657aa9b2f2126d37b).

vc_zoom/indico_vc_zoom/forms.py
moliholy: > I would also consider to not allow attaching zoom meeting to an event if registration is enabled

Yes, I agree with that. Done in [6b167dd](https://github.com/indico/indico-plugins/pull/292/commits/6b167dd47a559cff21a7fd1184bbd36044090a2f).

vc_zoom/indico_vc_zoom/plugin.py
moliholy: Heh, indeed. I am usually so focused on not making mistakes and if so spotting them quickly to get PRs in proper shape that I tend to forget what a non-failing workflow looks like 😆 .

Done in [a7361cd](https://github.com/indico/indico-plugins/pull/292/commits/a7361cddb7bd0238f10fc99645ae267426f4af81).

moliholy:
@ThiefMaster OK... I think that's all.

vc_zoom/README.md
```diff
@@ -11,6 +11,11 @@
 
 ## Changelog
 
+### 3.3.8
```
ThiefMaster: 3.3.7 has not been released yet, so it should go there. But actually I realized just now that neither has been 3.3.6 (which kind of makes sense, these small changes were not that important to justify a new plugin release).

So please move both this changelog entry and the language interpretation changelog entry to 3.3.6 and set `version = '3.3.6'` in pyproject.toml

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -42,11 +43,48 @@
...
+    return any(required_scopes <= granted_scopes for required_scopes in scope_options)
+
+
+def _get_missing_auto_registration_scopes(granted_scopes, *, allow_webinars):
+    missing_scopes = []
+    keys = ('meeting', 'webinar') if allow_webinars else ('meeting',)
+    for key in keys:
+        scope_options = AUTO_REGISTRATION_SCOPE_OPTIONS[key]
+        if not _has_required_zoom_scopes(granted_scopes, scope_options):
+            best_missing = min((required - granted_scopes for required in scope_options), key=len)
```
ThiefMaster: Can this result in legacy scopes being proposed?

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -665,3 +754,246 @@ def _event_metadata_postprocess(self, sender, event, data, user=None, skip_acces
...
+                else:
+                    room_ops[zoom_id]['remove'].add(email)
+        for ops in room_ops.values():
+            ops['add'] = list(ops['add'].values())
+        return room_ops
+
+    def _has_other_active_room_registration(self, vc_room, email, exclude_ids):
+        from indico.modules.events.models.events import Event
+        from indico.modules.events.registration.models.forms import RegistrationForm
+        from indico.modules.events.registration.models.registrations import Registration
```
ThiefMaster: These imports all look safe to have on the top level w/o introducing circular imports.

vc_zoom/tests/scope_test.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def test_auto_registration_scope_check_accepts_broad_webinar_scopes():
+    scopes = set(AUTO_REGISTRATION_MEETING_SCOPES) | set(AUTO_REGISTRATION_BROAD_WEBINAR_SCOPES)
+    assert _get_missing_auto_registration_scopes(scopes, allow_webinars=True) == ()
+
+
+def test_auto_registration_scope_check_requires_webinar_scopes_when_enabled():
+    # When no webinar scopes are present, the function reports the option set with fewest missing scopes
+    # (legacy: 2 scopes) rather than the modern set (3 scopes)
```
ThiefMaster: OK this seems to answer my question asked in another comment... I don't think it's a good idea to ever propose legacy scopes unless the app is currently already using legacy scopes.

vc_zoom/README.md
moliholy: OK, all merged into 3.3.6 in [705f141](https://github.com/indico/indico-plugins/pull/292/commits/705f141a90b356542f4390287033631b5861aad3).

vc_zoom/tests/scope_test.py
moliholy: Scopes controlled in [7ee10e6](https://github.com/indico/indico-plugins/pull/292/commits/7ee10e641721c1c2c2cc54aa2e56130686d7783b).

vc_zoom/indico_vc_zoom/plugin.py
moliholy: Fixed in [7ee10e6](https://github.com/indico/indico-plugins/pull/292/commits/7ee10e641721c1c2c2cc54aa2e56130686d7783b).

vc_zoom/indico_vc_zoom/plugin.py
moliholy: Indeed. Addressed in [7ee10e6](https://github.com/indico/indico-plugins/pull/292/commits/7ee10e641721c1c2c2cc54aa2e56130686d7783b).

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -665,3 +760,239 @@ def _event_metadata_postprocess(self, sender, event, data, user=None, skip_acces
...
+                'action': 'cancel',
+                'registrants': [{'id': rid, 'email': email}
+                                for email, rid in registrant_map.items()]
+            }
+            if is_webinar:
+                client.update_webinar_registrants_status(zoom_id, status_data)
+            else:
+                client.update_meeting_registrants_status(zoom_id, status_data)
+        except HTTPError:
+            self.logger.warning(f'Could not remove registrants from Zoom {zoom_type} %s',  # noqa: G004
```
ThiefMaster: Include the list of emails here

ThiefMaster: the registration id would also be useful (both here and when adding)

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -665,3 +760,239 @@ def _event_metadata_postprocess(self, sender, event, data, user=None, skip_acces
...
+
+    def _vc_room_created(self, vc_room, event, **kwargs):
+        if vc_room.type != self.service_name or not self.settings.get('allow_auto_register'):
+            return
+        if not vc_room.data.get('auto_register'):
+            return
+        for regform in event.registration_forms:
+            for registration in regform.active_registrations:
+                if registration.state == RegistrationState.complete:
+                    self._queue_registration_sync(registration, remove=False)
```
ThiefMaster, line 817: This also needs to be done when enabling the auto-registration option so existing registrants get added on Zoom.

moliholy: Done in [a5145e0](https://github.com/indico/indico-plugins/pull/292/commits/a5145e0dd626301e325080a17aeb44319bc56c76).

vc_zoom/indico_vc_zoom/plugin.py
moliholy: Done in [a5145e0](https://github.com/indico/indico-plugins/pull/292/commits/a5145e0dd626301e325080a17aeb44319bc56c76).

ThiefMaster: ah sorry, I meant the indico registration id, not the zoom registration id. having the latter as well isn't necessarily bad though

moliholy: Done in [9f581bf](https://github.com/indico/indico-plugins/pull/292/commits/9f581bf397ea980459b5d80bafe0aab69fb6e31e).

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -42,11 +45,52 @@
                                  process_alternative_hosts, update_zoom_meeting)
 
 
+AUTO_REGISTRATION_MEETING_SCOPES = ('meeting:read:list_registrants:admin', 'meeting:write:registrant:admin',
+                                    'meeting:update:registrant_status:admin')
```
ThiefMaster, line 51: found another missing scope: `meeting:write:batch_registrants:admin` (+ `webinar:write:batch_registrants:admin`)

moliholy: Good catch. Fixed in [c1dd39f](https://github.com/indico/indico-plugins/pull/292/commits/c1dd39fde423a3cf31fa84490696f7fd9bee87ef).

vc_zoom/README.md
```diff
@@ -11,12 +11,11 @@
 
 ## Changelog
 
-### 3.3.7
-
-- Support Zoom's [Language Interpretation](https://support.zoom.com/hc/en/article?id=zm_kb&sysparm_article=KB0064768) feature
-
 ### 3.3.6
 
+- Add automatic Zoom registration: event managers can opt-in to automatically register/unregister
```
ThiefMaster: How does this feature work for registrants that do not have an Indico account? Since Zoom does NOT send emails when you're registered, they do not have the Zoom link, and because they are not loggeg-in on Indico, it cannot show the zoom link for them in there either...

The only solutions I can think of are:

- Indico sending an email with the Zoom link when you are auto-registered
- Displaying the personal zoom links on the registration details page which you can access w/ your registration token (that's in the registration confirmation email)

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -318,6 +401,14 @@ def update_data_vc_room(self, vc_room, data, is_new=False):
 
         flag_modified(vc_room, 'data')
 
+        # If auto_register was just enabled on an existing room, sync existing registrants
+        if not is_new and not auto_register_before and vc_room.data.get('auto_register'):
+            for event_assoc in vc_room.events:
+                for regform in event_assoc.event.registration_forms:
+                    for registration in regform.active_registrations:
+                        if registration.state == RegistrationState.complete:
+                            self._queue_registration_sync(registration, remove=False)
```
ThiefMaster: This sends all the registrations - what about getting the current list of registrations from Zoom and only sending those that are not registered yet?

moliholy: Cleaner, yes. Done in [bd10f8e](https://github.com/indico/indico-plugins/pull/292/commits/bd10f8efb728530eb062914d8b1f5901e86d763b).

vc_zoom/README.md
moliholy: Hmmm yes, that's true. I think the simplest is to just display the personal zoom link in the registration details page.

On the other hand, emails can get lost, specially if the user is registering for an event in say +6 months.

So... I guess using both should cover most of the cases. Or perhaps only displaying the zoom link would suffice 🤔 

@ThiefMaster what's your take?

ThiefMaster: I'd go for the registration details page for now, and typically event organizers send emails closer to an event anyway, and those emails include the token to access the registration details anyway.

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -665,3 +787,245 @@ def _event_metadata_postprocess(self, sender, event, data, user=None, skip_acces
...
+    def _registration_created(self, registration, **kwargs):
+        self._queue_registration_sync(registration, remove=False)
+
+    def _registration_deleted(self, registration, **kwargs):
+        self._queue_registration_sync(registration, remove=True)
+
+    def _registration_state_updated(self, registration, **kwargs):
+        if registration.state == RegistrationState.complete:
+            self._queue_registration_sync(registration, remove=False)
+        elif registration.state == RegistrationState.withdrawn:
```
ThiefMaster:
```suggestion
        elif registration.state in {RegistrationState.pending, RegistrationState.withdrawn}:
```

otherwise "reset approval" doesn't remove the registration from zoom

PS: Please rebase

moliholy: Done in [d892766](https://github.com/indico/indico-plugins/pull/292/commits/d8927660e6e716222a95858425b294f290dab233). Rebased as well.

vc_zoom/README.md
moliholy: Done in [ba878db](https://github.com/indico/indico-plugins/pull/292/commits/ba878db28d6a1b2136574c8207d31ec9857d9c30).

Screenshot 2026-04-15 at 15 02 43

vc_zoom/README.md
```diff
@@ -15,6 +15,8 @@
 
 - Add automatic Zoom registration: event managers can opt-in to automatically register/unregister
   Indico participants in the corresponding Zoom meeting or webinar
+- Show the personalized Zoom join URL on the registration summary page so participants
+  without an Indico account can retrieve it from their registration token link
```
ThiefMaster: Please do not add multiple changelog entries for the same feature. I don't think this level of detail is needed here so I'd stick with the original changelog entry you added for this feature.

moliholy: Yeah, too much information. Removed in [94b1247](https://github.com/indico/indico-plugins/pull/292/commits/94b1247a83f7e6ab4f973f7f11436d5f39376bbd).

vc_zoom/indico_vc_zoom/templates/registration_zoom_link.html
```diff
@@ -0,0 +1,17 @@
...
+    <tr>
+        <th class="regform-done-caption">
+            {% if join_urls|length > 1 %}
+                {{ entry.name }}
+            {% else %}
+                {% trans %}Join link{% endtrans %}
+            {% endif %}
+        </th>
+        <td class="regform-done-data">
+            <a href="{{ entry.url }}" target="_blank" rel="noopener noreferrer" style="word-break: break-all;">{{ entry.url }}</a>
```
ThiefMaster: I don't think this looks great. I'd rather have the usual zoom join link style...

Image

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -263,6 +263,7 @@ def init(self):
         self.connect(signals.vc.vc_room_created, self._vc_room_created)
         self.connect(signals.core.after_process, self._flush_pending_registrations)
         self.template_hook('event-vc-room-list-item-labels', self._render_vc_room_labels)
+        self.template_hook('before-render-registration-info', self._render_registration_zoom_link)
```
ThiefMaster, line 267: This is not ideal but it looks like we do not have a better for it. IMHO the best place would be a new template hook in `render_registration_summary` after the `receipts-list` div, but of course this would not work w/ the current Indico version - so let's maybe just try to make it look a bit nice in the place where it's being added now.

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -862,6 +863,50 @@ def get_personalized_join_url(self, vc_room, event_vc_room, user):
             cached_urls[cache_key] = result
         return result
 
+    def _get_personalized_join_url_for_registration(self, vc_room, registration):
```
ThiefMaster: This looks quite similar to `get_personalized_join_url` - can the two be combined into a single function?!

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -862,6 +863,50 @@ def get_personalized_join_url(self, vc_room, event_vc_room, user):
...
+        if vc_room.type != self.service_name or not self.settings.get('allow_auto_register'):
+            return None
+        if not vc_room.data.get('auto_register'):
+            return None
+
+        cache_key = None
+        if has_request_context():
+            cache_key = (vc_room.id, registration.id)
+            cached_urls = g.setdefault('zoom_personalized_join_urls_by_registration', {})
+            if cache_key in cached_urls:
```
ThiefMaster: Do we get any cache hits on this in the end? Because if not, then maybe the cache logic should not be there in the first place... (also in the other method)

vc_zoom/indico_vc_zoom/templates/registration_zoom_link.html
moliholy: The idea was to show that the link was personalized, but yes... not ideal. Does this look better?

Screenshot 2026-04-15 at 16 49 05

ThiefMaster: You could add something like "(Personalized link)" behind it if you want to make it clearer...

But yes, it's better. The button like in other places would be even better IMHO. 
image

You could also try just rendering the Zoom heading and then not using two columns below but a single wider one...

moliholy: Attempted to address it in [2b1298f](https://github.com/indico/indico-plugins/pull/292/commits/2b1298f84381256d2e5c1a581517ba12b7c86ccc).

Screenshot 2026-04-15 at 17 22 43


Good enough?

vc_zoom/indico_vc_zoom/plugin.py
moliholy: I agree. Leaving the hook for now since moving it would not be ideal right now. Happy to address it in the future.

vc_zoom/indico_vc_zoom/plugin.py
moliholy: Yep! ~~[eb8ae80](https://github.com/indico/indico-plugins/pull/292/commits/eb8ae80258886b120b3ac3f3689db398166fbe2a)~~ [a49eb46](https://github.com/indico/indico-plugins/pull/292/commits/a49eb465391d69a77be979e4f750ca5fb434c3e1).

vc_zoom/indico_vc_zoom/plugin.py
moliholy: Within a single request the same (vc_room, user) pair does get resolved more than once (join button macro + info box popup, and now also the iCal export path), so the cache is saving duplicate Zoom API round-trips on those pages.

That said, if you'd prefer keeping the code simpler I can drop the `flask.g` caching and rely on Zoo being fast enough.

ThiefMaster: Yes then it makes sense to keep it. You could try simply using the `@memoize_request` on the method though - if that works it saves you some time.

moliholy: Done in [6de3664](https://github.com/indico/indico-plugins/pull/292/commits/6de36641fa44d96de7bf93bf1861d769564ac3c9).

vc_zoom/indico_vc_zoom/templates/registration_zoom_link.html
ThiefMaster: yeah, i can live with that ;)

vc_zoom/indico_vc_zoom/plugin.py
ThiefMaster: I think somehow this got reverted in 05d8a1ac9. I still don't like the idea of having this and no-one visibility enabled by default... it would be extremely intrusive e.g. on CERN Indico where most people do not need it.

moliholy: Alright [5104884](https://github.com/indico/indico-plugins/pull/292/commits/51048848542a706a09dddf41f2c9f0f84278d565).

ThiefMaster (approved):

PR #315, opened 2026-05-14
VC/Zoom: Push approval_type when auto_register toggles

moliholy:
This PR fixes a bug in Zoom registration:

**Before:** toggling `auto_register` on an existing room never pushed `settings.approval_type` to Zoom, so the meeting kept `approval_type=2` (registration disabled). The first registrant sync then failed with `Registration has not been enabled for this meeting`.

**After:** both `update_room` and `update_data_vc_room` now push `approval_type` (`0` when `auto_register` is on, `2` when off) before syncing registrants, so the initial backfill succeeds.

This PR also:
- Disables the `auto_register` toggle (with help text) when the linked event has already started, since Zoom silently drops `approval_type` on type=3/9 meetings without a future start time.
- Blocks room creation with `auto_register=True` on past events.
- Re-fetches the meeting after PATCH and flashes a warning if Zoom dropped the value.

### DIFF
--- a/vc_zoom/indico_vc_zoom/plugin.py
+++ b/vc_zoom/indico_vc_zoom/plugin.py
@@ -406,8 +406,17 @@ class ZoomPlugin(VCPluginMixin, IndicoPlugin):
 
         flag_modified(vc_room, 'data')
 
-        # If auto_register was just enabled on an existing room, sync existing registrants
+        # If auto_register was just enabled on an existing room, push approval_type to Zoom
+        # before syncing registrants so the initial backfill can hit the registrants endpoint.
         if not is_new and not auto_register_before and vc_room.data.get('auto_register'):
+            is_webinar = vc_room.data.get('meeting_type') == 'webinar'
+            if not is_webinar:
+                try:
+                    update_zoom_meeting(vc_room.data['zoom_id'],
+                                        {'settings': {'approval_type': 0}}, is_webinar=False)
+                except HTTPError:
+                    self.logger.warning('Could not enable registration on Zoom meeting %s',
+                                        vc_room.data['zoom_id'])
             candidates = [
                 registration
                 for event_assoc in vc_room.events
@@ -422,7 +431,7 @@ class ZoomPlugin(VCPluginMixin, IndicoPlugin):
                         self._find_zoom_registrants(ZoomIndicoClient(), vc_room, candidate_emails)
                     )
                 except HTTPError:
-                    zoom_type = 'webinar' if vc_room.data.get('meeting_type') == 'webinar' else 'meeting'
+                    zoom_type = 'webinar' if is_webinar else 'meeting'
                     self.logger.warning(f'Could not fetch registrants for Zoom {zoom_type} %s; '  # noqa: G004
                                         'syncing all existing registrants', vc_room.data['zoom_id'])
                     already_registered = set()
@@ -570,6 +579,9 @@ class ZoomPlugin(VCPluginMixin, IndicoPlugin):
                 changes.setdefault('settings', {})['participant_video'] = not vc_room.data['mute_participant_video']
             if vc_room.data['waiting_room'] != zoom_meeting_settings['waiting_room']:
                 changes.setdefault('settings', {})['waiting_room'] = vc_room.data['waiting_room']
+            desired_approval_type = 0 if vc_room.data.get('auto_register') else 2
+            if zoom_meeting_settings.get('approval_type') != desired_approval_type:
+                changes.setdefault('settings', {})['approval_type'] = desired_approval_type
 
         if changes:
             update_zoom_meeting(vc_room.data['zoom_id'], changes, is_webinar=is_webinar)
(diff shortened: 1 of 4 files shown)


moliholy:
@ThiefMaster pinging for visibility. This is a bug currently hitting production.

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -406,8 +406,17 @@ def update_data_vc_room(self, vc_room, data, is_new=False):
 
         flag_modified(vc_room, 'data')
 
-        # If auto_register was just enabled on an existing room, sync existing registrants
+        # If auto_register was just enabled on an existing room, push approval_type to Zoom
+        # before syncing registrants so the initial backfill can hit the registrants endpoint.
         if not is_new and not auto_register_before and vc_room.data.get('auto_register'):
+            is_webinar = vc_room.data.get('meeting_type') == 'webinar'
+            if not is_webinar:
```
ThiefMaster: webinars don't need this?

ThiefMaster: According to the Zoom API docs, `approval_type` exists for webinars as well. So I think you may need to handle it there as well.

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -406,8 +406,17 @@ def update_data_vc_room(self, vc_room, data, is_new=False):
...
+                try:
+                    update_zoom_meeting(vc_room.data['zoom_id'],
```
ThiefMaster: I tested this and it didn't work for me. The PATCH request succeeded, but the subsequent requests to get registrants or register participants failed with "Registration has not been enabled for this meeting". And when retrieving the meeting details from Zoom, settings.approval_type was still 2.

ThiefMaster: OK weird, even when creating anew zoom I'm no longer able to set approval_type 0... So maybe a problem on the Zoom side? Can you test if it's also broken for you and open a support case w/ zoom if it's a known issue?

vc_zoom/indico_vc_zoom/plugin.py
moliholy: Apologies, I had some issues with my Zoom account when testing this that pointed me to the wrong direction. But you're right, it's a similar pattern. It's fixed in [14e31fb](https://github.com/indico/indico-plugins/pull/315/commits/14e31fbd063298005de54e29feb85fa5292e092e).

vc_zoom/indico_vc_zoom/plugin.py
moliholy: It's working like a charm for me... it's just enabling/disabling the "Automatic registration" toggle and making sure the value gets persisted. Before changes in this PR whatever value was set upon creation could not be modified.

ThiefMaster: Weird. Must be a problem on Zoom's side that does not affect everyone... I assume you tested again today? And there were actually registrations that the plugin synced (or at least tried to sync)?

moliholy: I tested again literally 5 minutes ago, yes. However, it was in a meeting without users. Will test soon in a meeting with users and let you know.

moliholy: I think I know what happens... I think zoom silently drops the `approval_type` for recurring/no-fix-time meetings (`type=3`). It accepts it for `type=2` (scheduled).

moliholy: Yes, confirmed. Should we merge this and I make a follow-up PR, or all in this one?

ThiefMaster: Oh that explains it, my test event was in the past. I think checking it here would be nice.

moliholy: OK, done in https://github.com/indico/indico-plugins/pull/315/commits/5d337b529842d44a5e5fe76bcd8100c16525eecb. PR description updated with the changes. All yours.

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -581,21 +616,39 @@ def update_room(self, vc_room, event):
...
         zoom_meeting, is_really_webinar = fetch_zoom_meeting(vc_room)
         if is_webinar != is_really_webinar:
             vc_room.data['meeting_type'] = 'webinar' if is_really_webinar else 'regular'
         vc_room.name = zoom_meeting['topic']
+        zoom_approval_type = zoom_meeting['settings'].get('approval_type')
+        expected_approval_type = 0 if vc_room.data.get('auto_register') else 2
+        if zoom_approval_type != expected_approval_type and has_request_context():
+            flash(_('Zoom registration setting for this meeting differs from the Indico '
+                    'configuration. Save the videoconference to push the local setting to Zoom.'),
+                  'warning')
```
ThiefMaster: I don't think flashing messages here is a good idea, this gets called in places such as event cloning where it makes no sense to show this warning.

moliholy: Dropped the flash entirely in 9dc8ee525660c4f3fd8ded1806d93fa749d081a0. The save-time `logger.warning` in `update_room` remains as the detection path.

PR #320, opened 2026-05-28
VC/Zoom: Increase videoconference name limit to 200 characters

moliholy:
The videoconference name was capped at 60 characters, too short for many real meeting titles: customers ran into the limit with standard working-group names. Zoom's own meeting and webinar topic accepts up to 200 characters, so this raises the cap to match. Other videoconference backends keep the existing limit.

Companion to indico/indico#7560, which relaxes the shared 60-character limit in the core so each plugin can set its own service-specific cap.

### DIFF
--- a/vc_zoom/indico_vc_zoom/forms.py
+++ b/vc_zoom/indico_vc_zoom/forms.py
@@ -9,7 +9,7 @@ from flask import session
 from flask_pluginengine import current_plugin
 from wtforms.fields import BooleanField, StringField
 from wtforms.fields.simple import TextAreaField
-from wtforms.validators import DataRequired, ValidationError
+from wtforms.validators import DataRequired, Length, ValidationError
 
 from indico.modules.vc.forms import VCRoomAttachFormBase, VCRoomFormBase
 from indico.util.date_time import now_utc
@@ -64,6 +64,10 @@ class VCRoomForm(VCRoomFormBase):
     ]
 
     skip_fields = set(advanced_fields) | VCRoomFormBase.conditional_fields
+
+    # Override the core 60-char cap: Zoom's meeting/webinar `topic` accepts up to 200.
+    name = StringField(_('Name'), [DataRequired(), Length(min=3, max=200)],
+                       description=_('The name of the videoconference.'))
 
     meeting_type = IndicoRadioField(_('Meeting Type'),
                                     description=_('The type of Zoom meeting to be created'),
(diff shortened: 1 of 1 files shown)


ThiefMaster:
I would even be OK with removing that limit (or better making it much larger) in the core. Limiting it to what's supported by the actual VC service should be up to the plugin of course...

moliholy:
@ThiefMaster I can make another PR to the core, but unless we want to set the limit to 200 characters this should stay. Or even in that case IMO this should stay, as the other one could be modified in the future with side effects. What do you think?

ThiefMaster:
Yeah I think for now it's enough to make the change here.

vc_zoom/indico_vc_zoom/forms.py
```diff
@@ -65,6 +65,10 @@ class VCRoomForm(VCRoomFormBase):
 
     skip_fields = set(advanced_fields) | VCRoomFormBase.conditional_fields
 
+    # Override the core 60-char cap: Zoom's meeting/webinar `topic` accepts up to 200.
+    name = StringField(_('Name'), [DataRequired(), Length(min=3, max=200)],
+                       description=_('The name of the videoconference.'))
```
ThiefMaster:
```suggestion
                       description=_('The name of the Zoom meeting.'))
```

No need to be generic now that this lives in the zoom plugin

vc_zoom/indico_vc_zoom/forms.py
```diff
@@ -65,6 +65,10 @@ class VCRoomForm(VCRoomFormBase):
 
     skip_fields = set(advanced_fields) | VCRoomFormBase.conditional_fields
 
+    # Override the core 60-char cap: Zoom's meeting/webinar `topic` accepts up to 200.
+    name = StringField(_('Name'), [DataRequired(), Length(min=3, max=200)],
```
ThiefMaster, line 69: Please test to make sure the name shows up at the correct position in the form, I don't know if we manually specify an order somewhere or use the default iteration order of the form.

vc_zoom/indico_vc_zoom/forms.py
moliholy: Done in [ff333e3](https://github.com/indico/indico-plugins/pull/320/commits/ff333e3e470813bb63ef44064514873a6c38898f).

vc_zoom/indico_vc_zoom/forms.py
moliholy: I tested this:

Screenshot 2026-05-28 at 13 41 58
Screenshot 2026-05-29 at 11 14 55

Looks good.

PR #321, opened 2026-06-03
VC/Zoom: Send confirmation email to batch meeting registrants

moliholy:
Batch-added meeting registrants now get a Zoom confirmation email, matching single-registrant adds.

### DIFF
--- a/vc_zoom/indico_vc_zoom/plugin.py
+++ b/vc_zoom/indico_vc_zoom/plugin.py
@@ -1053,6 +1053,8 @@ class ZoomPlugin(VCPluginMixin, IndicoPlugin):
         for i in range(0, len(entries), BATCH_REGISTRANTS_MAX):
             chunk = entries[i:i + BATCH_REGISTRANTS_MAX]
             data = {'auto_approve': True, 'registrants': [e['data'] for e in chunk]}
+            if not is_webinar:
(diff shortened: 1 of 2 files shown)


moliholy:
@ThiefMaster for reference: https://developers.zoom.us/docs/api/meetings/#tag/invitation-registration/post/meetings/{meetingId}/batch_registrants

moliholy:
Webinars do not allow that parameter, as can be seen in the [docs](https://developers.zoom.us/docs/api/meetings/#tag/webinars/post/webinars/{webinarId}/batch_registrants). I tested and it does not crash if sending the parameter, but I'd prefer not to rely on the API not breaking, so better to skip sending that parameter if not needed.

ThiefMaster:
Does the parameter do anything for webinars? If it can send the email, then maybe passing it is good. If it's a no-op, then I think it's indeed better to not send it.

moliholy:
> Does the parameter do anything for webinars?

It is not documented in the API (for webinars! for meetings it IS documented), so the call would not follow the documentation. When I tried it simply ignored the parameter, but I didn't think it'd be good to rely on Zoom ignoring parameters... I know it makes the code a bit uglier though. Your call.

ThiefMaster:
Yes, I agree (like I wrote in my last comment)

PR #322, opened 2026-06-09
fix: require manual approval for auto_register Zoom meetings

moliholy:
Fixes an access-control hole in `auto_register` Zoom meetings.

These meetings were created with `approval_type=0` (automatic approval), so Zoom's public registration page approved anyone who self-registered, regardless of their Indico registration state (cancelled, rejected, or never registered). A leaked registration link let anyone obtain a join link and join the meeting, bypassing Indico as the gatekeeper.

Meetings now use `approval_type=1` (manual approval): self-registrants stay pending, while Indico's own registrants are still approved automatically via the `auto_approve` flag already sent on the registrant push, which Zoom only honors under `approval_type=1`.

### DIFF
--- a/vc_zoom/indico_vc_zoom/plugin.py
+++ b/vc_zoom/indico_vc_zoom/plugin.py
@@ -446,11 +446,13 @@ class ZoomPlugin(VCPluginMixin, IndicoPlugin):
 
         flag_modified(vc_room, 'data')
 
-        # Push approval_type whenever auto_register is toggled. On enable we also sync any
-        # existing registrants so the initial backfill can hit the registrants endpoint.
+        # Push approval_type whenever auto_register is toggled. Manual approval (1) keeps the public
+        # registration page gated (self-registrants stay "pending"); Indico-pushed registrants are
+        # auto-approved via auto_approve. On enable we also sync any existing registrants so the
+        # initial backfill can hit the registrants endpoint.
         if not is_new and auto_register_before != vc_room.data.get('auto_register'):
             is_webinar = vc_room.data.get('meeting_type') == 'webinar'
-            desired_approval_type = 0 if vc_room.data.get('auto_register') else 2
+            desired_approval_type = 1 if vc_room.data.get('auto_register') else 2
             self._push_approval_type(vc_room, desired_approval_type, is_webinar=is_webinar)
             if vc_room.data.get('auto_register'):
                 candidates = [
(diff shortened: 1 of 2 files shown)


vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -532,7 +534,12 @@ def create_room(self, vc_room, event):
                     'join_before_host': self.settings.get('join_before_host'),
                 })
                 if vc_room.data.get('auto_register'):
-                    settings['approval_type'] = 0
+                    # Manual approval (1), not automatic (0): the public Zoom registration page then
+                    # holds self-registrants in "pending", so a leaked link can't self-approve into
+                    # the meeting. Indico's own registrants are still approved automatically via the
+                    # auto_approve flag on the registrant push. Zoom only honors auto_approve when the
+                    # meeting was originally created with approval_type=1, hence setting it here.
```
ThiefMaster: What does this imply for meetings that have been created w/o auto-registration enabled where it was turned on later?

moliholy: That the bug prevails. Are you proposing some sort of "migration" of existing meetings?

ThiefMaster: No absolutely not (migration).

But what I meant are meetings created after merging this PR with registration disabled, where you then enable it after creation. The comment sounds like not everything will work then, unlike in a meeting that had registration enabled from the beginning...

moliholy: Oh no, I see you meant the "was originally created" part of the comment. That was the original feedback I received and I based the fix upon, but I just tested and it's actually valid for the current value of `approval_type`. New meetings created after this PR gets merged will be set to 1. Furthermore, other existing meetings will be set to 1 if they toggle the auto-register setting off/on (although I'd be careful with this, since if there are registrants they will receive a cancellation email!).

Comment updated in [5cbe063](https://github.com/indico/indico-plugins/pull/322/commits/5cbe063436e05abc8e6772dc298ee9df88af2d2c).

moliholy: > Furthermore, other existing meetings will be set to 1 if they toggle the auto-register setting off/on (although I'd be careful with this, since if there are registrants they will receive a cancellation email!)

Alternatively, they can change this setting to "Manually Approve" in the Zoom event's settings page. It won't be overwritten by Indico after the PR gets merged.

Screenshot 2026-06-09 at 23 38 29

ThiefMaster: I don't really care about existing meetings - especially since this isn't part of any release yet, and on our server it's not enabled, so besides you probably nobody is using the feature yet.

The updated comment is indeed much clearer :)
I'll test it tomorrow and merge it

ThiefMaster (approved):

PR #323, opened 2026-06-10
fix: handle already-removed room in the Zoom deletion webhook

moliholy:
Deleting a Zoom meeting or webinar in Indico occasionally returned an internal error and made Zoom keep retrying, because Indico had sometimes already removed the room by the time Zoom's deletion notification arrived. The notification now handles an already-removed room gracefully instead of erroring.


ThiefMaster:
Thanks for the proper fix - we actually get that error quite a bit as well, but had just muted it in sentry ;)

ThiefMaster (approved):

PR #328, opened 2026-06-30
VC/Zoom: Resolve registrant emails via a bulk user directory

moliholy:
When automatic registration is enabled on an event, resolving each registrant's Zoom-account email used to trigger its own `get_user` call. On an event with hundreds or thousands of registrants that meant one Zoom API call per person on every registration and approval, and once per participant on check-in.

This PR removes that per-registrant call on both fronts:

- **Bulk sync paths** (pushing registrants when the setting is turned on, and syncing on registration or approval) resolve everyone against a cached account user directory instead of asking Zoom per registrant. A daily background task walks the account once through List Users and caches the result, so no request ever pays for that walk. When the directory has not been cached yet, or a registrant is not in it, the lookup falls back to the per-user query.
- **Participant check-in** is resolved with a single indexed database lookup. Zoom already gives us the participant's exact email, so the matching registration is found directly (by registration email, account email, or identity, depending on the configured user lookup mode) with no Zoom call, keeping the webhook well within Zoom's response window.

On those same bulk paths, the check that decides whether a registrant is already covered by another active registration also used to run one database query per registrant. It now builds the set of active registrations for the event once per sync and reuses it, so a sync of thousands of registrants issues a single query instead of one per person.

It also fixes a related correctness issue on the sync path: a registration created while still awaiting moderation is no longer cancelled in Zoom. Only completed registrations are ever added to a meeting, so a pending one has nothing to cancel; the queued sync is now a no-op for it instead of issuing a cancellation that reached the registrant as a spurious cancellation notice.

Outgoing Zoom API requests are also logged at debug level, so the exact sequence of calls a registration or approval triggers can be traced when diagnosing a slow or misbehaving event.

Note: the directory lookup needs the `user:read:list_users:admin` scope, which is now part of the scope check that guards automatic registration.

### DIFF
--- a/vc_zoom/README.md
+++ b/vc_zoom/README.md
@@ -203,6 +203,7 @@ The scopes to select when creating the app are:
 - `meeting:write:batch_registrants:admin` (optional, only needed for automatic registration)
 - `meeting:update:registrant_status:admin` (optional, only needed for automatic registration)
 - `user:read:user:admin`
+- `user:read:list_users:admin` (optional, only needed for automatic registration)
 
 - `webinar:read:webinar:admin` (optional, only needed when using webinars)
 - `webinar:write:webinar:admin` (optional, only needed when using webinars)
--- a/vc_zoom/indico_vc_zoom/plugin.py
+++ b/vc_zoom/indico_vc_zoom/plugin.py
@@ -44,7 +44,7 @@ from indico_vc_zoom.notifications import notify_host_start_url
 from indico_vc_zoom.task import refresh_meetings
 from indico_vc_zoom.util import (UserLookupMode, ZoomMeetingType, fetch_zoom_meeting, find_enterprise_email,
                                  gen_random_passcode, get_alt_host_emails, get_schedule_args, get_url_data_args,
-                                 process_alternative_hosts, update_zoom_meeting)
+                                 preload_zoom_account_directory, process_alternative_hosts, update_zoom_meeting)
 
 
 AUTO_REGISTRATION_MEETING_SCOPES = ('meeting:read:list_registrants:admin', 'meeting:write:registrant:admin',
@@ -463,6 +463,7 @@ class ZoomPlugin(VCPluginMixin, IndicoPlugin):
                     if registration.state == RegistrationState.complete
                 ]
                 if candidates:
+                    preload_zoom_account_directory()
                     candidate_emails = {self._get_registrant_email(r) for r in candidates}
                     try:
                         already_registered = set(
--- a/vc_zoom/indico_vc_zoom/plugin.py
+++ b/vc_zoom/indico_vc_zoom/plugin.py
@@ -1012,9 +1012,15 @@ class ZoomPlugin(VCPluginMixin, IndicoPlugin):
             email = self._get_registrant_email(registration)
 
             for vc_room in zoom_rooms:
-                if should_add and self._has_other_active_room_registration(vc_room, email, pending_add_ids):
-                    continue
-                if not should_add and self._has_other_active_room_registration(vc_room, email, pending_remove_ids):
+                if should_add:
+                    if self._has_other_active_room_registration(vc_room, email, pending_add_ids):
+                        continue
+                elif remove:
+                    if self._has_other_active_room_registration(vc_room, email, pending_remove_ids):
+                        continue
+                else:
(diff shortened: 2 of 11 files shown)


moliholy:
@ThiefMaster this is a bug currently, as with thousands of users the request returns a 504 and:

- Leaves an inconsistent state: not all users are registered on Zoom, but a subset of them.
- Both auto-registration and check-in flags are not saved, as such operation is done after all participants are registered, but the operation gets aborted.

Regardless: it is very inefficient to perform O(n) queries to Zoom just get the emails.

moliholy:
I confirm the optimizations are working fine after the fix was deployed in a customer's infra a few days ago.

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -968,10 +974,19 @@ def _queue_registration_sync(self, registration, *, remove):
...
+        try:
+            preload_zoom_account_directory()
+        except HTTPError:
+            self.logger.warning('Could not preload the Zoom account directory; falling back to per-user lookups')
+
     def _flush_pending_registrations(self, sender, **kwargs):
         if not (pending := g.pop('zoom_pending_registrations', None)):
             return
 
+        self._preload_directory()
```
ThiefMaster: I think doing this unconditionally is extremely inefficient for the - often much more common - case of having few registrations and many Zoom users.

It would be good to have a setting for a threshold and/or cache the list of Zoom users in a more persistent way (for a few hours) and fall back to the individual lookup only for users that aren't in the cache.

moliholy: I added a hardcoded limit of 25 users to trigger the fetch in [b55f974](https://github.com/indico/indico-plugins/pull/328/commits/b55f97488a05a6057112365f945c8f9b4c1a4aee). I find a plugin setting too technical for users, as they don't really have to know about this issue nor I think can really evaluate the consequences.

In any case happy to change it if you think otherwise.

moliholy:
@ThiefMaster any news on this PR?

vc_zoom/indico_vc_zoom/plugin.py
ThiefMaster: Fine for now. I have to see how long it takes on a large tenant like ours to load the whole user mapping. In case it's too slow, I may ask to move this to a celery task so the cache is usually always populated...

vc_zoom/README.md
```diff
@@ -202,6 +202,7 @@ The scopes to select when creating the app are:
 - `meeting:write:batch_registrants:admin` (optional, only needed for automatic registration)
 - `meeting:update:registrant_status:admin` (optional, only needed for automatic registration)
 - `user:read:user:admin`
+- `user:read:list_users:admin` (optional, only needed for automatic registration)
```
ThiefMaster, line 210: Shouldn't this also be added to the list of permissions in `AUTO_REGISTRATION_MEETING_SCOPES` etc?

ThiefMaster:
Please fix the rebase conflicts (from merging #325)

vc_zoom/indico_vc_zoom/plugin.py
ThiefMaster: This needs to be done in Celery, even with our ~18k accounts this takes a whole minute.

```py
In [18]: %timeit -r1 -n1 preload_zoom_account_directory()
1min 12s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)

In [19]: len(preload_zoom_account_directory())
Out[19]: 17984
```

I'd suggest a daily task to fetch this data and cache it, and whenever the data is needed but not cached, fall back to querying it individually. Likewise for cache misses.

moliholy: Done in [4ec954b7](https://github.com/indico/indico-plugins/pull/328/commits/4ec954b7c9bd415e5bb90fe259503a27bb055fdc).

vc_zoom/README.md
moliholy: Done in [9f8bd03c](https://github.com/indico/indico-plugins/pull/328/commits/9f8bd03c3d75be2b2f74b6aa6892ffcfa6218e47).

moliholy:
@ThiefMaster ready for another round.

PR #337, opened 2026-08-10
VC/Zoom: Define alternative hosts on a meeting

moliholy:
This PR lets managers pick alternative hosts when creating or editing a Zoom meeting, so someone other than the host can start it.

Only Indico users with a Zoom account can be selected, and the meeting host cannot be one of them.

Screenshot 2026-08-10 at 17 07 08

### DIFF
--- a/vc_zoom/indico_vc_zoom/forms.py
+++ b/vc_zoom/indico_vc_zoom/forms.py
@@ -16,7 +16,7 @@ from indico.util.date_time import now_utc
 from indico.util.string import is_valid_mail
 from indico.util.user import principal_from_identifier
 from indico.web.forms.base import generated_data
-from indico.web.forms.fields import IndicoRadioField, MultipleItemsField, PrincipalField
+from indico.web.forms.fields import IndicoRadioField, MultipleItemsField, PrincipalField, PrincipalListField
 from indico.web.forms.util import inject_validators
 from indico.web.forms.validators import HiddenUnless, IndicoRegexp
 from indico.web.forms.widgets import SwitchWidget
@@ -81,6 +81,10 @@ class VCRoomForm(VCRoomFormBase):
 
     host_user = PrincipalField(_('User'),
                                [HiddenUnless('host_choice', 'someone_else'), DataRequired()])
+
+    alternative_host_users = PrincipalListField(_('Alternative hosts'),
+                                                description=_('These users can start the meeting in the absence of '
+                                                              'the host. They need a Zoom account as well.'))
 
     password = StringField(_('Passcode'),
                            [DataRequired(), IndicoRegexp(r'^\d{8,10}$')],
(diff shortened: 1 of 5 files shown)


ThiefMaster:
Did you disable maintainer pushes for this PR? I wanted to push the rebased version and add a changelog entry, but it failed...

vc_zoom/indico_vc_zoom/forms.py
```diff
@@ -221,6 +230,12 @@ def validate_host_user(self, field):
         if self.host_choice.data == 'someone_else':
             self._check_zoom_user(field.data)
 
+    def validate_alternative_host_users(self, field):
+        for user in (field.data or ()):
+            self._check_zoom_user(user)
```
ThiefMaster: The message from this field is quite generic which is perfectly fine (and good to avoid being unnecessarily verbose) for the Host field, but for the Co-host field this isn't ideal. It would be good to mention in the error which user doesn't have a Zoom account.

moliholy:
> Did you disable maintainer pushes for this PR?

Solved!

vc_zoom/indico_vc_zoom/forms.py
ThiefMaster: Also, can you test what happens if someone is co-host and then their Zoom account is removed from Zoom?

- The field must not break on load
- If saving fails until you remove that bad co-host it's fine (and actually expected)

Because right now (w/o your PR) we sometimes have this problem and it breaks in an obscure way because people don't really see why things fail and how to fix it...

moliholy: Hmmm this is going to be hard to test live. Let me see what I can do 🤔

moliholy: Done in [680e1e5](https://github.com/indico/indico-plugins/pull/337/commits/680e1e53c59847d42bf197a0666f112ed121175c), the error now names every user missing a Zoom account.

moliholy: Tested in [680e1e5](https://github.com/indico/indico-plugins/pull/337/commits/680e1e53c59847d42bf197a0666f112ed121175c): the field still loads them fine, and saving fails until the bad co-host is removed, now naming them. The same happens when the update is triggered outside the form (a reschedule, for instance), which was the obscure case since nobody sees the form there.

Still, I am figuring out whether I can reproduce it live, not just with unit tests. Problem is that my subscription does not allow adding more people to the organization. Let me see if I can figure that out.

vc_zoom/indico_vc_zoom/forms.py
```diff
@@ -250,6 +259,14 @@ def validate_host_user(self, field):
         if self.host_choice.data == 'someone_else':
             self._check_zoom_user(field.data)
 
+    def validate_alternative_host_users(self, field):
+        users = field.data or ()
+        # unlike the host field, this one holds several users, so the error has to name them
+        if invalid := sorted(user.full_name for user in users if find_enterprise_email(user) is None):
+            raise ValidationError(_('No Zoom account: {}').format(', '.join(invalid)))
```
ThiefMaster: Image

oops :D (already fixed)

ThiefMaster (approved):

PR #338, opened 2026-08-11
VC/Zoom: Let users choose the audio type of a meeting

moliholy:
This PR adds an `Audio` option to the advanced settings of a Zoom videoconference, so users can choose whether participants join through the telephone, computer audio, or both. New meetings keep whatever Zoom has as its own default, and the value stays in sync when the meeting is edited or refreshed.

Screenshot 2026-08-11 at 14 21 21

### DIFF
--- a/vc_zoom/indico_vc_zoom/fixtures.py
+++ b/vc_zoom/indico_vc_zoom/fixtures.py
@@ -89,6 +89,7 @@ def create_zoom_meeting(db, test_client, no_csrf_check, smtp, zoom_api):
                     'vc-mute_host_video': 'y',
                     'vc-mute_audio': 'y',
                     f'vc-{link_type}': obj.id,
+                    **{f'vc-{key}': value for key, value in kwargs.items()},
                 },
             )
             assert resp.status_code == 200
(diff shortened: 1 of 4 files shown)


ThiefMaster:
TBH I'm not sure about the usefulness of this...  what usecase do people have for changing this? AFAIK the default on Zoom is "both". The goal of the plugin was never to expose every single advanced setting from Zoom, considering that the zoom client can be used in the (likely rare?) cases where someone wants to change such settings...

moliholy:
It's a use case for our customer to enforce the way users engage in Zoom meetings. These settings can of course be tweaked (like any other though) via Zoom directly, but it'd make sense to cover the main ones related to the participants' degrees of freedom. There can be security reasons for this, as meetings can be sensitive and the way they are accessed can pose a risk.

OmeGak:
Is possible for participants to join the call before the host? If so, they may join via an unauthorized method before the host is able to configure the setting. Even then, not having this Zoom setting on Indico pushes the responsibility to the Zoom call host, which may not even be aware that this is a requirement.

ThiefMaster:
Zoom doesn't have a setting to restrict what can be chosen for a meeting (I guess the goal is to disable phone-in?)?

In the (probably very rare) case that an organization has a specific requirement there it sounds like this is something that should be limited on the Zoom tenant... unless of course they are fine (and actually want) to support phone-in for other events.

moliholy:
Ideally they'd like to disable phone-in in all events, but I thought it'd be too aggressive to have a plugin setting and instead opted for the per-event setting. Don't mind changing it to plugin setting if that makes more sense.

tomasr8:
> There can be security reasons for this, as meetings can be sensitive and the way they are accessed can pose a risk.

Out of curiosity, what kind of risks are we talking about?

moliholy:
> Out of curiosity, what kind of risks are we talking about?

Allowing ways to access a meeting for unauthorized people. If accessing via zoom itself (web browser or app) you require to be authenticated. Not the case for telephone access. Meetings can be extremely sensitive and it's important to restrict the access depending on the case.

@ThiefMaster as you mentioned the tenant can easily set this globally (see screenshot below). Let me get back to you once I clarify this 🙏 

Screenshot 2026-08-26 at 01 48 08

moliholy:
Closing this one: the audio type can be configured in the Zoom tenant settings, so exposing it per meeting makes no sense.

PR #345, opened 2026-09-15
VC/Zoom: Allow setting the secrets in indico.conf

moliholy:
This PR allows setting the Zoom credentials (account ID, client ID and client secret) and the webhook secret token in `indico.conf`. When set there they take precedence and the corresponding plugin settings get disabled; otherwise the settings page keeps working as before.

Requires indico/indico#7499.

Zoom plugin settings with both secret fields disabled


vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -185,22 +185,22 @@ class PluginSettingsForm(VCPluginSettingsFormBase):
     phone_link = URLField(_('Join via phone'), [Optional(), URL()],
                           description=_('Link to instructions on joining a meeting via phone'))
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        for name in ('client_secret', 'webhook_token'):
+            if getattr(ZoomPlugin.plugin_config, name.upper()):
+                self[name].render_kw = {'disabled': True}
```
ThiefMaster, line 192: I haven't tested it yet, but I didn't see anything in this code that hides the value... I think it would be good to do so (overwrite it with `*****`, `<value hidden>`, or something like that) since I don't think an admin should be able to view these secrets when they have been set via the config file.

moliholy: Good catch, the stored value is now masked so it never reaches the browser: [403059b4](https://github.com/indico/indico-plugins/pull/345/commits/403059b4edfbd3d73817a313189119d827c44581).

vc_zoom/indico_vc_zoom/plugin.py
```diff
@@ -239,6 +240,7 @@ class ZoomPlugin(VCPluginMixin, IndicoPlugin):
     vc_room_form = VCRoomForm
     vc_room_attach_form = VCRoomAttachForm
     friendly_name = 'Zoom'
+    plugin_config_defaults = {'CLIENT_SECRET': None, 'WEBHOOK_TOKEN': None}
```
ThiefMaster: Add account id and client id here as well. I don't think it makes sense that you can set half of the credentials via config and have to set the other half via web.

moliholy: Done in [207e6c10](https://github.com/indico/indico-plugins/pull/345/commits/207e6c10b43e5aa61af959a27ae077a80187faf2), all four values can now come from the config file.

ThiefMaster (approved):

